PR #3, opened 2019-07-12
feat(cache): starting to sketch out cache API

zkat:
This is an initial sketch of the stuff I'm working towards with the cache API. This is hard! I'm kinda blocked right now so I'm just putting up this draft request until such a time I can ask questions or even pair on the rest of it, mostly because I can't figure out how to do certain things with the async API, and because I don't understand why some of the code is failing to typecheck.


zkat:
I'm implementing this as a separate crate for now!

Issue #18, opened 2019-07-30
Middleware: follow redirects


yoshuawuyts:
We should write a middleware that automatically follows redirects.

derekdreery:
You might want to mention server-side request forgery in the docs for this middleware (https://github.com/seanmonstar/reqwest/issues/680).

yoshuawuyts:
What I think the first step to resolving this is, is to enable redirect support for GET requests without a body. That way we don't have any issues about trying to clone streams and payloads, and can safely retry without idempotency issues.

yonkeltron:
Thanks for all the hard word!

I'd love to throw my enthusiasm behind this issue as it just bit me. Would be happy to help test this functionality out with a pre-release version or something like that. Def willing to provide feedback.

Fishrock123:
https://github.com/http-rs/surf/issues/192 also addresses things related to this, namely potential middleware re-use and non-consumption of `surf::Request`.

yoshuawuyts:
Related discussion: https://github.com/algesten/ureq/issues/187

Fishrock123:
#195 should have closed this

PR #21, opened 2019-07-30
Follow redirects

yoshuawuyts:
Builds on #17. Attempt to implement #18. But ooph, this is not easy.

## parse url
A lot of `Location` headers are relative or even just a part of a URL. We should have something like node's `url.resolve()` to create complete URLs. Possibly as an extension to the url crate.

```js
const resolvedUrl = url.resolve(req.url, res.headers.get('location'))
let redirectURL = url.parse(resolvedUrl)

if (isURL.test(res.headers.get('location'))) {
redirectURL = url.parse(res.headers.get('location'))
```

## Detect streams
So if we're going to retry requests, we can't just be streaming from a moving target. Instead we need to make sure we own the streaming resource.

I think we may need to include more information in the middleware, which means changing the `client` param to a `Context` struct to contain more information. We can the follow make-fetch-happen's lead and add a `is-stream` flag to it, and not retry if it's not from a source that wont' mutate.

## Auth retries
I kind of don't want to touch this one for now; seems tricky hah.

## References
- https://github.com/zkat/make-fetch-happen/blob/latest/index.js#L329
- https://github.com/servo/rust-url/issues/2

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,24 +12,31 @@ readme = "README.md"
 edition = "2018"
 
 [features]
-default = ["middleware-logger"]
+default = ["chttp", "middleware-logger"]
+hyper-client = ["hyper", "runtime", "runtime-raw", "runtime-tokio" ]
+chttp-client = ["chttp"]
 middleware-logger = []
 
 [dependencies]
-hyper = { version = "0.12.32", default-features = false }
 http = "0.1.17"
 futures-preview = { version = "0.3.0-alpha.17", features = ["compat", "io-compat"] }
 serde = "1.0.97"
 serde_json = "1.0.40"
-hyper-tls = "0.3.2"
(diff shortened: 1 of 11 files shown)


jonathandturner:
@yoshuawuyts - any thoughts on how this might work?  I think we're pretty flexible. You probably know the redirect ins and outs better than I do :)

yoshuawuyts:
@jonathandturner heh, trying to read back what I wrote four months ago is tricky.

What I think the first step to resolving this is, is to enable redirect support for GET requests without a body. That way we don't have any issues about trying to clone streams and payloads, and can safely retry without idempotency issues.

I also suspect that would make the most progress towards https://github.com/nushell/nushell/issues/409, as the most common case of "query a url, follow the redirect" will be solved.

PR #34, opened 2019-08-15
Implement convenience methods for hyper-client

badboy:
## Description

Implement convenience methods for use with hyper-client, enable TLS by default for hyper-client and fix the build by removing a deprecated method.

## Motivation and Context

I'm running on macOS with the curl library not immediately available.
Hyper on the other hand works just fine.
However, most convenience things weren't implemented.

This PR changes multiple things, I'm happy to split them up if wanted.

## How Has This Been Tested?

`cargo test --no-default-features --features hyper-client`

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ ] I have read the [CONTRIBUTING](https://github.com/rust-net-web/tide/blob/master/.github/CONTRIBUTING.md) document.

### DIFF
--- a/src/one_off.rs
+++ b/src/one_off.rs
@@ -1,6 +1,9 @@
 #[cfg(feature = "native-client")]
 use super::http_client::native::NativeClient;
 
+#[cfg(feature = "hyper-client")]
+use super::http_client::hyper::HyperClient as NativeClient;
+
 use super::Request;
 
 /// Perform a one-off `GET` request.
--- a/src/request.rs
+++ b/src/request.rs
@@ -458,7 +458,7 @@ impl<C: HttpClient> Request<C> {
     /// # Ok(()) }
     /// ```
     pub fn body_file(mut self, path: impl AsRef<Path>) -> io::Result<Self> {
-        let mime = mime_guess::guess_mime_type(&path);
+        let mime = mime_guess::from_path(&path).first_or_octet_stream();
         let bytes = fs::read(path)?;
         *self.req.as_mut().unwrap().body_mut() = bytes.into();
         Ok(self.set_mime(mime))
--- a/src/request.rs
+++ b/src/request.rs
@@ -21,9 +21,11 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
 
+#[cfg(feature = "hyper-client")]
+use super::http_client::hyper::HyperClient;
 #[cfg(feature = "native-client")]
 use super::http_client::native::NativeClient;
-#[cfg(feature = "native-client")]
+#[cfg(any(feature = "hyper-client", feature = "native-client"))]
 use std::convert::TryFrom;
 
 /// An HTTP request, returns a `Response`.
@@ -63,6 +65,32 @@ impl Request<NativeClient> {
     /// ```
     pub fn new(method: http::Method, url: Url) -> Self {
         Self::with_client(method, url, NativeClient::new())
+    }
+}
+
+#[cfg(feature = "hyper-client")]
+impl Request<HyperClient> {
+    /// Create a new instance.
+    ///
+    /// This method is particularly useful when input URLs might be passed by third parties, and
+    /// you don't want to panic if they're malformed. If URLs are statically encoded, it might be
+    /// easier to use one of the shorthand methods instead.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # #![feature(async_await)]
+    /// # #[runtime::main]
+    /// # async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+    /// use surf::{http, url};
+    ///
+    /// let method = http::Method::GET;
+    /// let url = url::Url::parse("https://httpbin.org/get")?;
+    /// let string = surf::Request::new(method, url).recv_string().await?;
+    /// # Ok(()) }
+    /// ```
+    pub fn new(method: http::Method, url: Url) -> Self {
+        Self::with_client(method, url, HyperClient::new())
     }
 }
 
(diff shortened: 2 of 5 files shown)


src/one_off.rs
```diff
@@ -31,7 +34,7 @@ use super::Request;
 /// let string = surf::get("https://httpbin.org/get").recv_string().await?;
 /// # Ok(()) }
 /// ```
-pub fn get(uri: impl AsRef<str>) -> Request<NativeClient> {
+pub fn get(uri: impl AsRef<str>) -> Request<Client> {
```
yoshuawuyts: Because we already have a `Client` struct, introducing another similarly named client in the same range of APIs may be confusing. I don't see any _great_ options here, but perhaps even something like this may work:

```rust
#[cfg(feature = "hyper-client")]
use super::http_client::hyper::HyperClient as NativeClient;
```

Or alternatively we could use a different name (if there are any better suggestions).

badboy: How about `HttpClient`?

yoshuawuyts: That's the name of the trait we're planning to expose that's used by `HyperClient` etc. It'll be public through the `with_client` method.. Uhhh, maybe it's okay to leave the one-off methods for now and only expose it through the `Client` instance?

badboy: It would be a shame if the one-off methods don't work, especially because the defaults (= native-client) fail to compile on macOS, but it all does work with hyper.

I'm fine with having `HyperClient as NativeClient`.
The docs will correctly expand this anyway (and docs for all things at the same time doesn't seem easily possible anyway)

src/request.rs
```diff
@@ -602,6 +630,22 @@ impl<R: AsyncRead + Unpin + Send + 'static> TryFrom<http::Request<Box<R>>>
...
+impl<R: AsyncRead + Unpin + Send + 'static> TryFrom<http::Request<Box<R>>>
+    for Request<HyperClient>
+{
+    type Error = io::Error;
+
+    /// Converts an `http::Request` to a `surf::Request`.
+    fn try_from(http_request: http::Request<Box<R>>) -> io::Result<Self> {
+        let (parts, body) = http_request.into_parts();
+        let url = format!("{}", parts.uri);
+        let req = Self::new(parts.method, url.parse().unwrap());
```
berkus: shouldnt it be `url.parse()?` here instead of unwrap?

berkus:
The hyper-client seems to mess up the body by doing some decode/encode behind the back? Haven't figured out yet what exactly happens.

This looks like
```
        let body = dbg!(surf::get(url.clone()).recv_string().await?);
```

the printed (supposedly String) body looks like
```
[bin/src/main.rs:8] surf::get(url.clone()).recv_string().await? = "({:committer_name \"REDACTED\", :why \"github\", :parallel 1, :committer_email \"REDACTED\", :vcs_tag nil, :username \"REDACTED\", :start_time #joda/inst \"2019-08-22T18:22:56.812Z\", :build_num 7799, :stop_time nil, :author_date \"2019-08-22T21:14:18+03:00\", :dont-build nil, :build_time_millis nil, :outcome nil, :lifecycle :running, :workflows {:job-name \"REDACTED\", :job-id \"676c054c-9d58-4ffd-b0fc-21ff1cf85d86\", :workflow-id \"241dc90a-e7a2-45e5-b819-b93d868beead\", :workspace-id \"241dc90a-e7a2-45e5-b819-b93d868beead\", :upstream-job-ids [\"db885f9d-8f45-4909-8257-6c615a927d3a\" \"aab32e88-cc7f-4a6c-bada-6398e1cf2e21\"], :upstream-concurrency-map {:aab32e88-cc7f-4a6c-bada-6398e1cf2e21 [\"db885f9d-8f45-4909-8257-6c615a927d3a\"], :db885f9d-8f45-4909-8257-6c615a927d3a [\"aab32e88-cc7f-4a6c-bada-6398e1cf2e21\"]}, :workflow-name \"REDACTED\"}, :vcs_revision \"b40839a941d218431e04802f282d238116ebedd4\", :build_url \"REDACTED\", :pull_requests ({:head_sha \"b40839a941d218431e04802f282d238116ebedd4\", :url \"REDACTED\"}), :fleet :picard, :vcs_url \"REDACTED\", :status :running, :author_name \"REDACTED\", :author_email \"REDACTED\", :queued-at #joda/inst \"2019-08-22T18:22:53.798Z\", :committer_date \"2019-08-22T21:14:18+03:00\", :branch \"REDACTED\", :reponame \"REDACTED\", :body \"\", :user {:is-user true, :login \"REDACTED\", :avatar-url \"https://avatars3.githubusercontent.com/u/REDACTED?v=4\", :name \"REDACTED\", :vcs-type :github, :id REDACTED}, :subject \"REDACTED\", :platform \"2.0\", :usage-queued-at #joda/inst \"2019-08-22T18:22:53.711Z\"} {:committer_name \"REDACTED\", :why \"github\", :parallel 1, :committer_email \"REDACTED\", :vcs_tag nil, :username \"REDACTED\", :start_time #joda/inst \"2019-08-22T18:22:58.727Z\", :build_num 7798, :stop_time nil, :author_date \"2019-08-22T21:14:18+03:00\", :dont-build nil, :build_time_millis nil, :outcome nil, :lifecycle :running,
```
(I REDACTED some private data out)

Looks like it went through some deserialize/serialize sequence, but where and how - not sure. Maybe hyper does some custom parsing if body has Content-Type set?
The `#joda/inst` tag looks suspicious though, as Joda is a Java library - ??.

I built hyper client using patches from this pull request and current master of surf.

yoshuawuyts:
We're in the process of moving our backend impls over to https://github.com/http-rs/http-client/. We should pick this PR back up there. Going to go ahead and close it for now. Thanks!

PR #43, opened 2019-08-18
Add support for hyper-rustls

### LINKED ISSUE #40: Switch to Rustls?
[Rustls](https://crates.io/crates/rustls) is an SSL/TLS library, written completely in Rust. It seems to be [more safe and fast than OpenSSL](https://www.zdnet.com/article/a-rust-based-tls-library-outperformed-openssl-in-almost-every-category/) (which is used by [native-tls](https://crates.io/crates/native-tls) on Linux).

Surf uses [native-tls](https://crates.io/crates/native-tls) if I'm not mistaken, which uses a native implementation on each supported platform. Would Rustls be more preferred to use in Surf?

wezm:
## Description

Allow the hyper client to be configured to use `hyper-rustls` for TLS abilities, eliminating the dependency on the OS TLS library/OpenSSL.

## Motivation and Context

Fixes #40 

This is a quick pass at using `hyper-rustls` instead of `native-tls`. `rustls` makes cross-compiling easier and eliminates a dependency on a C library (OpenSSL) on open source OSes.

## How Has This Been Tested?

These changes build with `cargo build --no-default-features --features hyper-client,hyper-rustls,rustls`. However, `cargo test --no-default-features --features hyper-client,hyper-rustls,rustls` fails. It appears that running the tests is broken on master with `native-tls` too, as `cargo test --no-default-features --features hyper-client,hyper-tls,native-tls` also fails.

## Types of changes

## Checklist:

- [x] I have read the [CONTRIBUTING](https://github.com/rust-net-web/tide/blob/master/.github/CONTRIBUTING.md) document.
- [ ] All new and existing tests passed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,9 @@ isahc = { version = "0.7", optional = true, default-features = false, features =
 # hyper-client
 hyper = { version = "0.12.32", optional = true, default-features = false }
 hyper-tls = { version = "0.3.2", optional = true }
+hyper-rustls = { version = "0.17.0", optional = true }
 native-tls = { version = "0.2.2", optional = true }
+rustls = { version = "0.16.0", optional = true }
 runtime = { version = "0.3.0-alpha.6", optional = true }
 runtime-raw = { version = "0.3.0-alpha.4", optional = true }
 runtime-tokio = { version = "0.3.0-alpha.5", optional = true }
--- a/src/http_client/hyper.rs
+++ b/src/http_client/hyper.rs
@@ -6,7 +6,13 @@ use futures::future::FutureObj;
 use futures::prelude::*;
 use futures::task::SpawnError;
 use hyper::client::connect as hyper_connect;
+#[cfg(feature = "hyper-tls")]
 use hyper_tls::HttpsConnector;
+#[cfg(feature = "hyper-rustls")]
+use hyper_rustls::HttpsConnector;
+#[cfg(feature = "hyper-rustls")]
+use rustls::ClientConfig;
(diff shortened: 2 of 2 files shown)


yoshuawuyts:
@wezm I'm kind of wondering if we shouldn't straight up replace hyper's tls impl with `rustls`. There's quite a few options possible for hyper and if possible I'd like to prevent a combinatorial explosion in our configs for hyper features. How do you feel about only exposing `rustls` as Hyper's only TLS impl?

wezm:
> How do you feel about only exposing rustls as Hyper's only TLS impl?

That would be fine with me.

Gymmasssorla:
> @wezm I'm kind of wondering if we shouldn't straight up replace hyper's tls impl with `rustls`. There's quite a few options possible for hyper and if possible I'd like to prevent a combinatorial explosion in our configs for hyper features. How do you feel about only exposing `rustls` as Hyper's only TLS impl?

I think that not all users of Surf may want to use Rustls and forget about a native TLS implementation (see the discussion in https://github.com/rustasync/surf/issues/40).

yoshuawuyts:
@Gymmasssorla We should probably make `http-client` public so any kind of configuration can be created.

yoshuawuyts:
[http-client](https://github.com/http-rs/http-client/) is public now, and lives as part of a separate crate. We should pick the work in this patch back up there. It's currently being integrated back into Surf; so fingers crossed we can add support for this.

Going to go ahead and close this patch for now, as we can't merge it into this repo anymore. Thanks heaps @wezm for your work!

PR #49, opened 2019-08-19
cHTTP has been renamed to isahc

vorot93:


berkus:
This seems to work and pulls entire body.

The hyper-client seems to mess up the body by doing some decode/encode behind the back? Haven't figured out yet what exactly happens.

Cargo.toml
```diff
@@ -15,7 +15,7 @@ edition = "2018"
 default = ["native-client", "middleware-logger"]
 native-client = ["curl-client", "wasm-client"]
 hyper-client = ["hyper", "runtime", "runtime-raw", "runtime-tokio" ]
-curl-client = ["chttp"]
+curl-client = ["isahc"]
```
yoshuawuyts: This is a breaking change we can't merge until we're comfortable releasing a 2.0. I feel like we may need to revert this and log it so we don't forget to update it at a later time (:

yoshuawuyts: Ah nevermind, I totally misread. We're good here!

CarlosLanderas:
How can we fix apps depending on surf version 1.0.1 with chttp if you can't merge this because is a breaking change? Thanks!

vorot93:
This PR has been merged and is there to stay

CarlosLanderas:
Is there any plan to release a new version? I've seen the last one is 1.0.1 and it doest not have the chttp read body fix right? Thanks

PR #51, opened 2019-08-21
Only require the http2 feature from chttp.

badboy:
Default features includes always statically linking curl, which fails under macOS.
Not requiring static linking does the right thing on all OSs:

* Link against a system library if it exists (e.g. on Linux and macOS)
* Statically build and link against curl if it doesn't (e.g. on Windows)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ url = "2.0.0"
 
 # chttp-client
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-chttp = { version = "0.5.3", optional = true }
+chttp = { version = "0.5.3", optional = true, default-features = false, features = ["http2"] }
 
 # hyper-client
 hyper = { version = "0.12.32", optional = true, default-features = false }


yoshuawuyts:
@sagebind this might be of interest for isahc too as it probably means this is breaking for direct users of isahc too unless they opt-out of static binding (where rust's `libcurl` bindings will always still fallback to static binding if no local copy is found).

sagebind:
Could you perhaps share some more details on what errors you are getting on macOS? Compiling statically passes on my macOS CI servers, and it also works fine on my Macbook.

> Not requiring static linking does the right thing on all OSs:

I think it could be up to debate as to what the "right thing" would be. :wink: Generally static linking is preferred unless there's a good reason not to do it, because it means you are running against a consistent, immutable version of the library and have less system dependencies; both of which are great for portability and testability.

badboy:
There are more details in here: https://github.com/alexcrichton/curl-rust/issues/279
And a PR is up for fixing static linking on macOS as well: https://github.com/alexcrichton/curl-rust/pull/283

sagebind:
This problem has now been fixed in the [0.9.0](https://github.com/sagebind/isahc/releases/tag/0.9.0) release, so if you're interested in static linking everywhere, it works now. :wink:

PR #58, opened 2019-08-25
Fix hyper client feature flag

jbcden:
Fix build when using the `hyper-client` feature flag.

## Description
This change adds some missing optional dependencies that needed to be included when using the `hyper-client` feature. This also updates some other packages that were bringing in a conflicting version of `futures-timer` when combined with the `hyper-client` feature flag.

## Motivation and Context
Any project using this library cannot build with the `hyper-feature`.

## How Has This Been Tested?
    You can replicate this issue with the following:

    1. Create a new rust project
    2. Add `surf` as a dependency with the `hyper-client` feature enabled
    3. Build the project.

    Instead of building successfully, you will see issues due to missing
    dependencies; mainly `hyper-tls` and `native-tls`.

After this change, this builds as expected.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] I have read the [CONTRIBUTING](https://github.com/rust-net-web/tide/blob/master/.github/CONTRIBUTING.md) document.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,9 +38,9 @@ isahc = { version = "0.7", optional = true }
 hyper = { version = "0.12.32", optional = true, default-features = false }
 hyper-tls = { version = "0.3.2", optional = true }
 native-tls = { version = "0.2.2", optional = true }
-runtime = { version = "0.3.0-alpha.6", optional = true }
-runtime-raw = { version = "0.3.0-alpha.4", optional = true }
-runtime-tokio = { version = "0.3.0-alpha.5", optional = true }
+runtime = { version = "0.3.0-alpha.7", optional = true }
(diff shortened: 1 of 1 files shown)


berkus:
Doesn't #34 fix the same, together with some compilation issues? It doesn't update async libs versions only.

jbcden:
@berkus Good call, I missed that! I think upgrading the async libs will be necessary as you get a compilation error without them though.

jbcden:
@berkus I just tested and it actually does work without the version bumps! I'm not exactly sure why that conflict is not happening now but I can't reproduce on that branch so it should be good!

berkus:
@jbcden yes, I have compiled hyper support successfully with #34 but it messes up my json bodies so it's not fully operational for me.

PR #61, opened 2019-08-26
Typed headers support

vorot93:
# What is this?
This PR introduces support for hyper 0.11-style typed headers. They allow for strong typing of all common HTTP headers and their values.

# Implementation note
While I prefer that we expose `HeaderMap` directly, this plumbing brings typed headers support to the `Headers` wrapper directly.

In the future we could completely obsolete stringly typed headers thanks to the extensibility provided by the `Header` trait.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ middleware-logger = []
 
 [dependencies]
 futures-preview = { version = "0.3.0-alpha.18", features = ["compat", "io-compat"] }
+headers = "0.2"
 http = "0.1.17"
 log = { version = "0.4.7", features = ["kv_unstable"] }
 mime = "0.3.13"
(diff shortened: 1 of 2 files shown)


vorot93:
It is really one of the four options:
- Expose `http::HeaderMap` and let end users or 3rd party crates use, which you rejected in the previous PR.
- Forward the calls to internal `http::HeaderMap` which is what this PR is about.
- Reimplement (copy-paste) the encoding logic
- Use `std::mem::transmute` to force expose `http::HeaderMap` and use ecosystem standard headers API.

1 and 2 already rejected. And I don't know what's worse: unsafe or half-arsed copy-and-paste.

yoshuawuyts:
My preferred solution here would be is to have header constructors that output both a "header name" + "header value", which can in turn be passed to our API as strings. I think there are other approaches possible too, but this has my strong preference we try out first.

Whether this should be a new crate, or an `Into` impl for `typed-headers` requires some deliberation -- but it's probably worth experimenting with different approaches.

Issue #28, opened 2019-08-08
http_client: change `AsyncRead` bound to `AsyncBufRead`


yoshuawuyts:
As per @fairingrey's suggestion, we should probably follow up on https://github.com/rustasync/http-service/issues/31 and use [`AsyncBufRead`](https://docs.rs/futures-preview/0.3.0-alpha.17/futures/io/trait.AsyncBufRead.html) to prevent double buffering.

PR #65, opened 2019-09-04
[WIP] Add compression middleware

### LINKED ISSUE #24: Middleware: compression
Using https://docs.rs/async-compression/0.1.0-alpha.1/async_compression/ we should write a middleware that sends `Accept-Encoding` parameters to the server, and decompresses the response.

## Examples
- https://github.com/rustasync/tide/blob/master/tide-compression/src/lib.rs

Nehliin:
## Description
The middleware inserts an `Accepted_Encoding` header with `"gzip, br, deflate, zstd"` as value and then decodes the response from the server depending on the`Content-Encoding`. The implementation is _heavily_ inspired of the [`Decompression` implemented in tide](https://github.com/rustasync/tide/blob/master/tide-compression/src/lib.rs). To implement a similar approach for surf I wrapped the `Body` used in requests and responses in a `BufReader` and used the different `AsyncBufRead` decoders from `async_compression`. 

## Motivation and Context
Fixes #24 

## How Has This Been Tested?
I have added a StubClient so the middleware can be tested with integration tests. More tested are needed though, for example responses with multiple layers of compression.

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have read the [CONTRIBUTING](https://github.com/rustasync/tide/blob/master/.github/CONTRIBUTING.md) document.
- [X] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,9 @@ wasm-client = ["js-sys", "web-sys", "wasm-bindgen", "wasm-bindgen-futures"]
 middleware-logger = []
 
 [dependencies]
+async-compression = {version = "0.1.0-alpha.1"}
+bytes = "0.4.12"
+accept-encoding = "0.2.0-alpha.2"
 futures-preview = { version = "0.3.0-alpha.17", features = ["compat", "io-compat"] }
 http = "0.1.17"
 log = { version = "0.4.7", features = ["kv_unstable"] }
@@ -68,5 +71,8 @@ features = [
 
 [dev-dependencies]
 femme = "1.1.0"
+flate2 = "1.0.11"
+brotli2 = "0.3.2"
+libzstd = { version = "0.4", package = "zstd", default-features = false }
 runtime = "0.3.0-alpha.6"
 serde = { version = "1.0.97", features = ["derive"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,10 +18,10 @@ hyper-client = ["hyper", "runtime", "runtime-raw", "runtime-tokio" ]
 curl-client = ["chttp"]
 wasm-client = ["js-sys", "web-sys", "wasm-bindgen", "wasm-bindgen-futures"]
 middleware-logger = []
+middleware-compression = ["async-compression"]
 
 [dependencies]
-async-compression = {version = "0.1.0-alpha.1"}
-bytes = "0.4.12"
+async-compression = { git = "https://github.com/Nemo157/async-compression-rs", branch = "0.17-override", feature="bufread", optional = true}
 accept-encoding = "0.2.0-alpha.2"
 futures-preview = { version = "0.3.0-alpha.17", features = ["compat", "io-compat"] }
 http = "0.1.17"
@@ -35,7 +35,7 @@ url = "2.0.0"
 
 # chttp-client
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-chttp = { version = "0.5.3", optional = true }
+chttp = { version = "0.5.5", optional = true }
 
 # hyper-client
 hyper = { version = "0.12.32", optional = true, default-features = false }
@@ -76,3 +76,4 @@ brotli2 = "0.3.2"
 libzstd = { version = "0.4", package = "zstd", default-features = false }
 runtime = "0.3.0-alpha.6"
 serde = { version = "1.0.97", features = ["derive"] }
+bytes = "0.4.12"
--- a/src/http_client/mod.rs
+++ b/src/http_client/mod.rs
@@ -7,6 +7,8 @@ use std::fmt::{self, Debug};
 use std::io;
 use std::pin::Pin;
 use std::task::{Context, Poll};
+use futures::{Stream};
+use bytes::Bytes;
 
 #[cfg(all(feature = "hyper-client", not(target_arch = "wasm32")))]
 pub(crate) mod hyper;
@@ -72,6 +74,27 @@ impl Body {
         }
     }
 }
+
+impl Stream for Body {
+    type Item = io::Result<Bytes>;
+
+    fn poll_next(self: Pin<&mut Self>,
+                  cx: &mut Context<'_>,) -> Poll<Option<Self::Item>> {
+        let mut buf = [0;1];
+        match self.poll_read(cx, &mut buf) {
+            Poll::Ready(Ok(n)) => {
+                if n == 0 {
+                    Poll::Ready(None)
+                } else {
+                    Poll::Ready(Some(Ok(Bytes::from(&buf[..]))))
+                }
+            },
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Err(e)) => Poll::Ready(Some(Err(e)))
+        }
+    }
+}
+
 
 impl AsyncRead for Body {
     #[allow(missing_doc_code_examples)]
--- a/src/http_client/mod.rs
+++ b/src/http_client/mod.rs
@@ -2,13 +2,13 @@
 use futures::future::BoxFuture;
 use futures::io::AsyncRead;
 
+use bytes::Bytes;
+use futures::Stream;
 use std::error::Error;
 use std::fmt::{self, Debug};
 use std::io;
 use std::pin::Pin;
 use std::task::{Context, Poll};
-use futures::{Stream};
-use bytes::Bytes;
 
 #[cfg(all(feature = "hyper-client", not(target_arch = "wasm32")))]
 pub(crate) mod hyper;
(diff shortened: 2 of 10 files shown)


src/http_client/mod.rs
```diff
@@ -73,6 +75,25 @@ impl Body {
...
+                } else {
+                    Poll::Ready(Some(Ok(Bytes::from(&buf[..]))))
+                }
+            }
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Err(e)) => Poll::Ready(Some(Err(e))),
+        }
+    }
+}
+
```
yoshuawuyts: Ah, so we should probably use the `AsyncRead` variant in `async-compression`. This might be related to #28 so the body can be read without an extra `memcpy`, but I *definitely* do not want to expose `bytes` in any public API.

Cargo.toml
```diff
@@ -20,6 +20,9 @@ wasm-client = ["js-sys", "web-sys", "wasm-bindgen", "wasm-bindgen-futures"]
 middleware-logger = []
 
 [dependencies]
+async-compression = {version = "0.1.0-alpha.1"}
+bytes = "0.4.12"
+accept-encoding = "0.2.0-alpha.2"
```
yoshuawuyts: We probably should make the dependencies part of a `middleware-compression` feature.

src/http_client/mod.rs
Nehliin: I think it is, I don't think we will need bytes if `AsyncBufRead` is used.

Nehliin:
Thank you! I presume you mean the `AsyncBufRead` version?

yoshuawuyts:
@Nehliin haa, yes I did!

Nehliin:
I'm not sure if I am misreading the documentation for async_compression but it seems like only the Deflate and Zlib _encoders_ are available to use with `AsyncBufRead`. Which makes it hard for me to do anything until support for decoders and or other compression algorithms are available. I also tried to experiment with changing the `reader` in the `Body` to use `AsyncBufRead` but had a hard time getting anywhere. I presume that would be out of scope for this PR and another one would need to be made for that. 

What would you suggest the next steps to be? Should I wait for more work on `async_compression` and wait until the `Body`  could make use of `AsyncBufRead` or should I try to clean up the current implementation so `Bytes` aren't exposed to the users? I'm not sure if that's possible though since the Decoders are dependent of it. Maybe some internal wrapper type could help with that but I'm not sure.

Nemo157:
@Nehliin just published `async-compression 0.1.0-alpha.5` with full encoder/decoder support for `AsyncBufRead`

Nehliin:
Great thanks! Hopefully I can have some new commits done this weekend :)

Nehliin:
@Nemo157 I can't use the latest version since that depends on `futures-preview v0.3.0-alpha.18` and `chttp` (latest version) has set the same dependency to `version = "=0.3.0-alpha.17"` so cargo can't find a compatible version :( Unfortunately the `chttp` crate has changed name to `isahc` so we would need to change out that dependency. But when I tried to replace the old with `isach`I ran in to other issues like compilation errors in `futures-timer v0.2.1` and even if I specify a newer version in `Cargo.toml` I end up with the same issue so some other dependency seems to depend on `0.2.1`. I'm a bit unsure how to fix this :confused:

Nemo157:
As a workaround until chttp/isahc can be updated you could use this as your dependency, it's 0.1.0-alpha.5 but with the dependency changed back to futures alpha.17:

```toml
async-compression = { git = "https://github.com/Nemo157/async-compression-rs", branch = "0.17-override" }
```

Nehliin:
That's great! Thanks for the quick reply :)

Nehliin:
I updated the code so it now uses the `AsyncBufRead` version of `async-compression` and added the `middleware-compression` feature. I also updated the tests to include some "onion layered" compression that the middleware should handle. Those tests currently fails but I think it's because I don't set the header correctly in the mocked response from the stub client. That should hopefully be an easy fix that I'll do in the beginning of next week.

Nehliin:
I have now updated documentation a bit and added errors, the integration tests pass but the doc tests that tests a request to [https://httpbin.org/gzip](https://httpbin.org/#/Response_formats/get_gzip) fails. with the error `Error: Custom { kind: InvalidData, error: "Invalid gzip header" }` which confuses me. It's like the body of the response is not encoded in a valid way. But I've tried to curl it and then decode the response with `GzDecoder` directly in a separate test used when I tried to debug and that works. My code only extracts the body and then wraps the body in to a `BufReader` as previously suggested and gives that to the `GzipDecoder` from `async-compression`.

I've been stuck on this for awhile do you have any suggestions on things I could try?

Nemo157:
When you say it decoded ok using `GzDecoder` are you referring to the synchronous one or the one from async-compression? It could be an async-compression bug, `flate2` doesn't expose the gzip header parsing code so we had to reimplement that, and it's definitely not 100% complete.

Nehliin:
Yes I am using the [`flate2::bufread::GzDecoder`](https://docs.rs/flate2/1.0.11/flate2/bufread/struct.GzDecoder.html). Ah then it's probably a bug as you said, makes sense. I should have said something earlier and perhaps save me some confusion :smile:. I found this if that may be of interest https://docs.rs/gzip-header/0.3.0/gzip_header/

Nemo157:
So, [throwing some debugging in](https://github.com/Nemo157/surf/commit/68b3d9276454fb6895d21e4edcd4c4d3c6b8f63a) I found this:

```
> cargo test --doc  --features middleware-compression -- middleware::compression
[...]
[src/middleware/compression/compression.rs:23] (type_name::<T>(), reader.poll_read(cx, buf)) = (
    "surf::http_client::Body",
    Ready(
        Ok(
            285,
        ),
    ),
)
[src/middleware/compression/compression.rs:24] std::str::from_utf8(buf) = Ok(
    "{\n  \"gzipped\": true, \n  \"headers\": {\n    \"Accept\": \"*/*\", \n    \"Accept-Encoding\": \"gzip, br, deflate, zstd\", \n    \"Content-Length\": \"0\", \n    \"Host\": \"httpbin.org\", \n    \"User-Agent\": \"curl/7.65.1-DEV chttp/0.5.5\"\n  }, \n  \"method\": \"GET\", \n  \"origin\": \"92.116.115.71, 92.116.115.71\"\n}\n\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u
[...]
```

that's before the data is passed into the `GzipDecoder`, so it appears the underlying body implementation is already gzip decoding the content, and it's correctly failing when attempting to double-decode it.

Nehliin:
Oh, well that explains it. It turns out that [isahc already decodes](https://github.com/sagebind/isahc/blob/8944f82803fa151be9abf0658046ca792f3886d4/tests/encoding.rs) automatically. Looks like hyper doesn't do any compression reading through this [discussion](https://github.com/hyperium/hyper/issues/2) I wanted to try it myself but couldn't because of a compiler error with the latest surf release which isn't that nice:

`error[E0277]: the trait bound ext::TimeoutStream<S>: futures_core::stream::Stream is not satisfied
   --> /home/oskar/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-timer-0.2.1/src/ext.rs:170:9
    |
170 | impl<S> TryStream for TimeoutStream<S>
    |         ^^^^^^^^^ the trait futures_core::stream::Stream is not implemented for ext::TimeoutStream<S>`

And if I use `1.01` or `1.0.0` I get undeclared module `hyper_tls` together with `native_tls` because I think someone forgot to add those dependencies to the `hyper-client` feature in Cargo.toml

Nehliin:
Maybe this PR isn't necessary now that isach already does this under the hood? Should I close the PR? I learned a lot either way so I'm happy anyways :)

yoshuawuyts:
> Maybe this PR isn't necessary now that isach already does this under the hood

Apologies for the slow reply. I do think this would be fantastic to have, even if Isahc already does this. Not all client backends support compression, and having an easy way to enable it is probably a win for everyone involved!

I'm not sure if *now* is the right time to pick this back up, since #131 will likely change the signature of our middleware, but overall this would be fantastic to have!

Fishrock123:
Hi @Nehliin, if you still want to persue this, could you open a new PR with something that is more like https://github.com/Fishrock123/tide-compress?

Client decompression is likely a necessary feature for Surf to switch to using async-h1 as the default backend: https://github.com/http-rs/surf/issues/217

Nehliin:
@Fishrock123 I won't have time to work on this for a while :/

PR #66, opened 2019-09-06
Makes http_client contents public.

### LINKED ISSUE #54: What's the type of `surf::Client::new()` ?
Hello, I'm trying to create a client instance and pass it to functions for reuse:
```
let client = surf::Client::new();
get_item(client, url);
```

But what type should I use for the function parameter ?
```
async fn get_item(client: ???, url: String) -> ...
```

IslandUsurper:
## Description

However, to retain stability guarantees, their implementations are kept private to the crate.

## Motivation and Context
Fixes #54.

`Client` and `Request` are hard to pass around to functions because of their type parameter. Every user of _surf_ has to make these functions generic over the chosen HTTP client, even though only one should ever be used at a time. By exposing the names of the HTTP clients, _surf_ users can use the specific type instead of a trait-bound parameter.

## How Has This Been Tested?

`cargo +nightly test` passed on Linux x64. I am unsure how to get the hyper-client and wasm-client features to pass tests.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

`IsahcClient` is no longer `Default`, but this didn't seem to be used anyway.

## Checklist:

### DIFF
--- a/src/http_client/isahc.rs
+++ b/src/http_client/isahc.rs
@@ -10,15 +10,9 @@ pub struct IsahcClient {
     client: Arc<isahc::HttpClient>,
 }
 
-impl Default for IsahcClient {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl IsahcClient {
     /// Create a new instance.
-    pub fn new() -> Self {
+    pub(crate) fn new() -> Self {
         Self {
             client: Arc::new(isahc::HttpClient::new().unwrap()),
         }
--- a/src/http_client/isahc.rs
+++ b/src/http_client/isahc.rs
@@ -1,3 +1,5 @@
+//! HTTP Client adapter for Isahc.
+
 use super::{Body, HttpClient, Request, Response};
 
 use futures::future::BoxFuture;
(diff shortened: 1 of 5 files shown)


yoshuawuyts:
The build errored. I've gone ahead and restarted it. -- we should merge this if it passes and cut a new minor release.

IslandUsurper:
The build timed out either during or after creating the sccache directory. When I ran the commands on my own computer, I discovered warnings for missing documentation, so I added some. Hopefully, Travis will work correctly for this new commit.

pimeys:
Would it be possible to not make the types public, but instead use dynamic dispatch in the `Client`? It would not be that big of an issue performance-wise, but would allow us to not have a type parameter in `Client` at all...

pimeys:
Wrote another issue about my suggestion https://github.com/rustasync/surf/issues/69

IslandUsurper:
Since #70 fixed the Travis cache, can we retry this build?

yoshuawuyts:
We've published https://github.com/http-rs/http-client today which makes this public. We should move Surf over to use it :sparkles:

Issue #67, opened 2019-09-09
How to get the request body


CarlosLanderas:
Hello guys. I am checking source code inside surf::Request and I am trying to find a way to inspect body bytes / body string without moving ownership because I need to send some signed headers with the content, and the use the same request to be sent using http client.

The method I'm trying to do would be something like this:

```rust

 pub(crate) async fn sign(request: &surf::Request<impl HttpClient>) {

        let host = request.url();
        let verb = request.method()
                        .to_string()
                        .to_uppercase();

      //Need to read request without moving

   }
```
Im seeing there is a bunch of methods to set the body but I can't find anyone to read the bytes that are already inside the body. Any way to do this?

Thanks!

k-nasa:
surely. I will try to add.

yoshuawuyts:
@CarlosLanderas hi, thanks for filing this! -- I think the answer is that there's not really a good way to go about this. Because when reading a request body, we *have* to assume it's being consumed. E.g. if we have a stream from a socket, or a circular buffer, after reading the data it can be lost again.

So instead what I'd recommend is to invert the logic. Sign the file inside `surf` middleware, as data goes through. If that's not possible yet, then we should put in some effort to support that use case.

Also apologies for the late reply; I've had a lot on my plate recently. Thanks heaps for your patience!

CarlosLanderas:
Thanks @yoshuawuyts

I acomplished this, I don't remember how because it was some time ago. :P

https://github.com/CarlosLanderas/azure-app-configuration/blob/master/src/client.rs
https://github.com/CarlosLanderas/azure-app-configuration/blob/master/src/request_sign.rs

PR #78, opened 2019-10-04
Add method to get the request body

k-nasa:
## Description

close: https://github.com/rustasync/surf/issues/67

Currently there seems to be no method to get the request body.
So I added it.

## Types of changes

## Checklist:

- [x] I have added tests to cover my changes.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -558,6 +558,11 @@ impl<C: HttpClient> Request<C> {
     pub fn request(&self) -> Option<&http_client::Request> {
         self.req.as_ref()
     }
+
+    /// Get a mutable HTTP request
+    pub fn request_mut(&mut self) -> Option<&mut http_client::Request> {
+        self.req.as_mut()
+    }
 }
 
 impl<C: HttpClient> Future for Request<C> {
--- a/src/request.rs
+++ b/src/request.rs
@@ -2,8 +2,8 @@ use futures::future::BoxFuture;
 use futures::prelude::*;
 use http::Method;
 use mime::Mime;
-use serde::Serialize;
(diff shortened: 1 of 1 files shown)


k-nasa:
How about this?

src/request.rs
```diff
@@ -558,6 +626,11 @@ impl<C: HttpClient> Request<C> {
     pub fn request(&self) -> Option<&http_client::Request> {
         self.req.as_ref()
     }
+
+    /// Get a mutable HTTP request
+    pub fn request_mut(&mut self) -> Option<&mut http_client::Request> {
+        self.req.as_mut()
+    }
```
yoshuawuyts: We probably don't want to expose this as part of our public API.

src/request.rs
```diff
@@ -469,6 +470,73 @@ impl<C: HttpClient> Request<C> {
...
+    /// # Ok(()) }
+    /// ```
+    pub async fn read_body_to_json<T: DeserializeOwned>(&mut self) -> Result<T, Exception> {
+        let req = self.request_mut().unwrap();
+
+        let mut bytes = Vec::new();
+        req.body_mut().read_to_end(&mut bytes).await?;
+
+        Ok(serde_json::from_slice(&bytes)?)
+    }
```
yoshuawuyts: isn't this the same as the `body_json` method?

yoshuawuyts:
@k-nasa I don't think we can do this because we have to assume data cannot be read multiple times, which means "inspecting" data would be equal to consuming it. I've written in more detail about it here: https://github.com/http-rs/surf/issues/67#issuecomment-554434153.

PR #83, opened 2019-10-19
Remove runtime and use async-std

AZanellato:
## Description
The runtime library is no longer being updated, so I changed it to `async-std`.

## Motivation and Context

When trying to do the github-sdk, I got into a lot of troubles when starting with the crate. I was trying to follow other examples but they all used the `runtime` library. 

## How Has This Been Tested?

I have changed the code to fit with what I understand to be the way this would be done using `async_std`, using `tasks`

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] I have added tests to cover my changes.

### DIFF
--- a/examples/hello_world.rs
+++ b/examples/hello_world.rs
@@ -1,12 +1,12 @@
-type Exception = Box<dyn std::error::Error + Send + Sync + 'static>;
+use async_std::task;
 
-#[runtime::main]
-async fn main() -> Result<(), Exception> {
-    femme::start(log::LevelFilter::Info)?;
+fn main()  {
+    femme::start(log::LevelFilter::Info);
 
-    let uri = "https://httpbin.org/get";
-    let string = surf::get(uri).recv_string().await?;
-    println!("{}", string);
-
-    Ok(())
+    task::block_on(async {
+        let uri = "https://httpbin.org/get";
+        let string = surf::get(uri).recv_string().await;
+        println!("{}", string.unwrap());
+    });
+    ()
 }
--- a/examples/hello_world.rs
+++ b/examples/hello_world.rs
@@ -1,12 +1,15 @@
-type Exception = Box<dyn std::error::Error + Send + Sync + 'static>;
 use async_std::task;
 
-fn main()  {
+// The need for Ok with turbofish is explained here
+// https://rust-lang.github.io/async-book/07_workarounds/03_err_in_async_blocks.html
+fn main() -> Result<(), surf::Exception> {
     femme::start(log::LevelFilter::Info);
 
     task::block_on(async {
         let uri = "https://httpbin.org/get";
-        let string = surf::get(uri).recv_string().await;
-        println!("{}", string.unwrap());
-    })
+        let string: String = surf::get(uri).recv_string().await?;
+        println!("{}", string);
+        Ok::<(), surf::Exception>(())
+    });
+    Ok(())
 }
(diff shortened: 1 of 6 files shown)


examples/hello_world.rs
```diff
@@ -1,12 +1,12 @@
...
-    let string = surf::get(uri).recv_string().await?;
-    println!("{}", string);
-
-    Ok(())
+    task::block_on(async {
+        let uri = "https://httpbin.org/get";
+        let string = surf::get(uri).recv_string().await;
+        println!("{}", string.unwrap());
+    });
+    ()
```
stjepang: Let's delete this line and the semicolon before it. The main function can still return `Result<(), Exception>` and we don't need the `.unwrap()` inside the `async` block.

AZanellato: @stjepang `task::block_on` seems to always return `()`. I didn`t found a way of bubbling up the return :(

AZanellato: I can remove the two lines and still maintain the `unwrap()` though :smile:

examples/next_reuse.rs
```diff
@@ -40,15 +41,15 @@ impl<C: HttpClient> Middleware<C> for Doubler {
...
-    femme::start(log::LevelFilter::Info)?;
-    let mut res = surf::get("https://httpbin.org/get")
-        .middleware(Doubler {})
-        .await?;
-    dbg!(&res);
-    let body = res.body_bytes().await?;
-    let body = String::from_utf8_lossy(&body);
-    println!("{}", body);
-    Ok(())
+fn main() {
```
yoshuawuyts: This should probably return:
```suggestion
fn main() -> Result<(), surf::Exception> {
```

so that `unwrap` isn't needed.

examples/middleware.rs
```diff
@@ -19,11 +19,14 @@ impl<C: HttpClient> Middleware<C> for Printer {
...
-#[runtime::main]
-async fn main() -> Result<(), surf::Exception> {
-    femme::start(log::LevelFilter::Info)?;
-    surf::get("https://httpbin.org/get")
-        .middleware(Printer {})
-        .await?;
-    Ok(())
+use async_std::task;
+
+fn main() {
```
yoshuawuyts: This should probably return:
```suggestion
fn main() -> Result<(), surf::Exception> {
```

so that `unwrap` isn't needed.

examples/hello_world.rs
```diff
@@ -1,12 +1,12 @@
 type Exception = Box<dyn std::error::Error + Send + Sync + 'static>;
+use async_std::task;
 
-#[runtime::main]
-async fn main() -> Result<(), Exception> {
-    femme::start(log::LevelFilter::Info)?;
+fn main()  {
```
yoshuawuyts: This should probably return:
```suggestion
fn main() -> Result<(), surf::Exception> {
```

so that `unwrap` isn't needed.

examples/next_reuse.rs
AZanellato: hey @yoshuawuyts , I tried this but the `unwrap` is inside the async block and I did't find any way to make that return an `Result`. :cry: 

How can I do that?

lnicola:
I think there's also the README example that needs to be updated.

yoshuawuyts:
@lnicola you're not wrong, but I propose we address that in a follow-up PR

PR #87, opened 2019-11-03
setup github actions

yoshuawuyts:
The migration to the http-rs org disabled CI. This sets up github actions for use here. Thanks!


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,86 @@
...
+      uses: actions-rs/cargo@v1
+      with:
+        command: check
+        args: --all --bins --examples
+
+    - name: check unstable
+      uses: actions-rs/cargo@v1
+      with:
+        command:  check
+        args: --all --benches --bins --examples --tests
```
k-nasa: Is Coco necessary?
I don't think there is a need to separate 'check' from 'check unstable'.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,86 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
```
k-nasa: Do you need master?

yoshuawuyts: For Bors the "trying" branch is usually enough. Perhaps we should add master too though.

.github/workflows/ci.yaml
yoshuawuyts: Oh yes, great point. I copied this in accidentally. Thanks heaps!

PR #90, opened 2019-11-07
1.0.3

yoshuawuyts:
Prepare release for 1.0.3. Currently `surf` is broken. This does not update us to `futures 0.3.0` since that's a breaking change. There are a few other breaking changes we to do in tandem with that. Thanks!


yoshuawuyts:
Unsure why cargo check isn't building, but merging anyway. Thanks!

PR #95, opened 2019-11-08
Remove broken link

mgrachev:


yoshuawuyts:
Heh, we've published that module today. You were right in filing this patch; but I guess it doesn't make sense to merge it anymore. Thanks for being attentive; hope this makes sense!

PR #99, opened 2019-11-11
fix readme example build error

uvd:

PR #105, opened 2019-11-18
return results in examples expecting results

### LINKED ISSUE #104: README examples do not return Result
Three of the four README examples have a `;` popped on the end of the `task...` block, so the function currently just returns `()` instead of a `Result`

foggy1:
Fixes #104


foggy1:
Just saw #99 ! closing this and the related issue

Issue #101, opened 2019-11-13
Response method body_string may use wrong character encoding


Hawk777:
The `body_string` method in the `Response` class is as follows:
```rust
    pub async fn body_string(&mut self) -> Result<String, Exception> {
        let bytes = self.body_bytes().await?;
        Ok(String::from_utf8(bytes).map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?)
    }
```

This just grabs the body as a byte array, and then decodes it with UTF-8. If it wasn’t originally encoded in UTF-8, this is wrong. What I think most people would expect `body_string` to do (I certainly did) is extract the character encoding from the `charset` parameter of the `Content-Type` header and decode the bytes to a string using that encoding.

While there is a note that “If the body cannot be interpreted as valid UTF-8, an `Err` is returned.” I found this confusing. I didn’t think it meant that the declared encoding was *ignored*, I thought it actually meant if the body *could not* be represented in UTF-8 (perhaps because it was binary data not matching any, or the declared, encoding). I am not sure whether this method, as currently written, is really all that useful? I found it rather misleading myself.

yoshuawuyts:
@Hawk777 Thanks for filing this; I hear you.

I think I built it this way because Rust strings are always guaranteed to be valid utf8. But if the encoding of the incoming request is different we should probably convert it first. Does that sound like it would be the right way to go?

Additionally we should probably indeed update the docs with more details of what we're doing. I hadn't really thought of it, but now that you mention it it's not surprising it can be confusing to people. Apologies!

Hawk777:
Yes, that is exactly what I would expect. It might also be worth deciding and documenting what the behaviour is (panic? return error?) if the response body is not actually text (e.g. `application/octet-stream` or some other non-textual type).

goto-bus-stop:
I think an `Err()` is the correct way to go if the encoding the header claims is unsupported or is incorrect. In that case the Err() should contain the byte array so you can decide to do something else, like passing it around as a byte array for `octet-stream`, or using a crate like [`chardet`](https://crates.io/crates/chardet/) to try to fall back an encoding that _does_ work.

[`encoding_rs`](https://crates.io/crates/encoding_rs) is very nice—perhaps it could be an optional dependency, so folks talking to an HTTP API that they know will return utf8 do not need to ship its conversion tables.

goto-bus-stop:
Should be addressed by #108! It tries to use the encoding from the `content-type` and returns an Err() if that doesn't work.

Issue #86, opened 2019-10-30
surf::Exception does not implement std::error::Error


dtolnay:
This is an unfriendly choice for errors returned by a library because it makes them not work with `?`.

```rust
async fn repro() -> Result<(), failure::Error> {
    let _ = surf::get("https://www.rust-lang.org").await?; // doesn't work
    Ok(())
}
```

```rust
async fn repro() -> anyhow::Result<()> {
    let _ = surf::get("https://www.rust-lang.org").await?; // doesn't work
    Ok(())
}
```

Application-focused error types like failure::Error are built on `impl<T: std::error::Error> From<T>` which is why it matters that library errors implement std::error::Error.

yoshuawuyts:
@dtolnay thanks for pointing this out; I agree this is not right, and we should fix this. Thanks for reporting, and sorry this spilled over into your issue tracker!

matklad:
@yoshuawuyts this might want to be marked with 2.0 milestone.

derekdreery:
From `anyhow::Error` docs

> Error works a lot like Box<dyn std::error::Error>, but with these differences:
> - Error requires that the error is Send, Sync, and 'static.
> - Error guarantees that a backtrace is available, even if the underlying error type does not provide one.
> - Error is represented as a narrow pointer — exactly one word in size instead of two.

 - The first one of these is fixed by using `Box<dyn std::error::Error + Send + Sync>` (I'm informed boxed errors are `'static` unless they have a lifetime parameter).
 - `anyhow::Error` creates a backtrace where the error is first converted into an `anyhow::Error`, I'm not sure that adds much in this use case over not having one.
 - `anyhow::Error` is a thin pointer — so you add one indirection to save one usize width. Again I don't think this matters much in this context: your `Result<T>` will be sized by `T` (unless it is less than 2 `usize`s), and unless you've only got a tiny amount of memory (in which case you can't use `alloc`, and therefore this crate) the nature of making http requests means that you're not going to create enough to make memory usage an issue.

All in all just use `Box<dyn std::error::Error + Send + Sync>` which is also `'static`. 1 less dependency and does the job of a type-erased error.

I made an RFC for [adding the type alias to std](https://github.com/rust-lang/rfcs/pull/2820).

dtolnay:
@derekdreery `Box<dyn Error + Send + Sync>` does not implement std::error::Error so it does not address this issue at all.

matklad:
[...]

PR #108, opened 2019-11-20
Use appropriate charset in `body_string()`

goto-bus-stop:
This is an initial attempt at addressing #101. Opening for comments—feel free to critique the direction etc and tell me to throw it all out if it's not what you're looking for! Leaving a few checkboxes…

A new default feature `"encoding"` opts in to using the appropriate character encoding.
- In native binaries, `encoding_rs` is used. If you know you're only talking to an API that always uses utf-8, you can turn off this feature and save the binary size that is taken up by the encoding tables.
- In WebAssembly, it calls out to the `TextDecoder` API. AFAIK This always makes a copy, so you can turn it off if you don't want to copy big strings between JS and wasm.
  - [x] We could optimize this for the utf-8 case by using `String::from_utf8()` directly if the charset is utf-8, and only use TextDecoder for other less common  encodings.

Turning off the `"encoding"` feature reverts to the current behaviour, which is assuming utf-8.
- [x] This should likely check that the charset is utf-8, and refuse to decode it if it is anything else, to avoid garbage output.

This patch also assumes utf-8 if no charset is specified in the response headers.
- [x] By spec the default encoding should be iso-8859-1, but in practice it's not. Need to figure out the best approach here.

This patch introduces an error type that's quite unlike anything Surf has right now, so I'm not sure if that direction fits in with the design. I also wrapped the new error type in an `io::ErrorKind::InvalidData` everywhere which is mostly cargo culting, I don't think that's actually necessary? I carried it over from the old `body_string()` implementation, where it _did_ make sense.
- [x] probably get rid of that wrapping

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,12 +12,13 @@ readme = "README.md"
 edition = "2018"
 
 [features]
-default = ["native-client", "middleware-logger"]
+default = ["native-client", "middleware-logger", "encoding"]
 native-client = ["curl-client", "wasm-client"]
 hyper-client = ["hyper", "runtime", "runtime-raw", "runtime-tokio" ]
 curl-client = ["isahc"]
 wasm-client = ["js-sys", "web-sys", "wasm-bindgen", "wasm-bindgen-futures"]
 middleware-logger = []
+encoding = ["encoding_rs"]
 
 [dependencies]
 futures-preview = { version = "0.3.0-alpha.19", features = ["compat", "io-compat"] }
@@ -30,8 +31,11 @@ serde_json = "1.0.40"
 serde_urlencoded = "0.6.1"
 url = "2.0.0"
 
-# isahc-client
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+# encoding
+encoding_rs = { version = "0.8.20", optional = true }
+
+# isahc-client
 isahc = { version = "0.7", optional = true, default-features = false, features = ["http2"]  }
 
 # hyper-client
@@ -63,6 +67,7 @@ features = [
     "RequestMode",
     "RequestRedirect",
     "Response",
+    "TextDecoder",
     "Window",
 ]
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,7 +88,7 @@ pub use url;
 
 pub use client::Client;
 pub use request::Request;
-pub use response::Response;
+pub use response::{DecodeError, Response};
 
 #[cfg(feature = "native-client")]
 mod one_off;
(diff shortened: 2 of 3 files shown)


goto-bus-stop:
Found a website that doesn't use UTF-8 in the year of our lord: https://shop-pro.jp/

Unfortunately, it still doesn't work with this PR, because it has a bunch of bytes that probably _are_ UTF-8 or some other encoding…

Another fallback strategy that folks could use would be to use encoding_rs to just decode it with the detected encoding anyways and accept that there will be UTF-8 replacement characters in the output, like Firefox does:
image

I've had some trouble finding good examples… a lot of sites that use non-UTF-8 encodings just don't add a `charset=` parameter to the mime type. This one uses the shift-jis encoding _and_ identifies itself as such: https://wowma.jp/ On master, `body_string()` returns a UTF-8 error, and with this patch it works, so that's good :sparkles: 

For comparison, the cURL CLI doesn't handle encoding at all, so both of the above sites "work" but output garbage if you're using a terminal emulator that uses utf8.

goto-bus-stop:
I'm not sure how to get the `DecodeError` out of the current `surf::Exception` type, since I think `downcast()` only exists on `Box<dyn Any>`? So maybe the `DecodeError` type is not useful yet, but if surf switches to `anyhow` I think its downcast() method would work ok.

CryZe:
> but if surf switches to anyhow I think its downcast() method would work ok.

anyhow is not the right abstraction for libraries, thiserror / snafu are.

goto-bus-stop:
So thinking about the above I think this is probably a good start as is:
- For pages that contain incorrect bytes, like the shop-pro.jp site, failing is likely better than silently producing lots of UTF8 replacement characters. Typically you'll use surf for APIs and not full web pages, which are more likely to contain incorrect bytes. There could be a `body_string_lossy()` method in the future for those use cases. Users who need to be super robust and deal with sites that use uncommon encodings but do not specify an encoding in `Content-Type` can manually check the content-type beforehand and decide to do character type detection at that point

- The DecodeError problem will likely be addressed by the solution to #86 
- Since the vast vast majority of sites uses UTF-8, it is likely a better default than ISO-8859-1 would be

yoshuawuyts:
> anyhow is not the right abstraction for libraries

@CryZe it's the right fit here though, as middleware can return arbitrary errors which are up to the user, and we don't want to enforce a generic param to capture any specific error.

---

@goto-bus-stop the summary you provided makes sense! Thanks for checking this against real-world examples; agree with utf-8 being a sensible default. Thank you!

PR #111, opened 2019-11-23
Futures 0.3

### LINKED ISSUE #100: update to futures 0.3
Now that async/await is stable, futures 0.3 and async-std 1.0 is released, it would be nice to be able to use surf with all these new stable versions of those projects.

goto-bus-stop:
isahc 0.8.0 is out :tada: 

- Update futures-related dependencies to ones that use v0.3
- Use the #[main] and #[test] attributes from async-std instead of from runtime
- Add missing dependencies to the `hyper-client` feature

I didn't update hyper to the alpha that uses futures 0.3 yet. That could be a later PR.

There is still an issue with the hyper implementation, it doesn't compile. This is also true on the main branch right now. I started porting it to async-std, which helped a bit, but it seems there's no way to actually use the HyperClient anyway right now and designing a way to access it seems out of scope for this PR. I guess the http_client module is moving into its own crate anyways?

Fixes #100

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,7 +72,6 @@ features = [
 ]
 
 [dev-dependencies]
-async-std = "1.0"
+async-std = { version = "1.0", features = ["attributes"] }
 femme = "1.1.0"
-runtime = "0.3.0-alpha.8"
 serde = { version = "1.0.97", features = ["derive"] }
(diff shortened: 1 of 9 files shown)


yoshuawuyts:
@goto-bus-stop yeah, I think it's best if we focus on curl + wasm for now. Filed a patch in https://github.com/http-rs/http-client/issues/5 so we can port that to in order to set us up for the Surf 2.0.0 release.

Fantastic work as usual; thanks so much for this!

PR #113, opened 2019-11-24
Convert `Exception` to `Error`.

### LINKED ISSUE #86: surf::Exception does not implement std::error::Error

derekdreery:
And use `anyhow::Error` rather than `Box<dyn Error+...>`.

fixes #86 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ quick script, or a cross-platform SDK, Surf will make it work.
 ```rust
 use async_std::task;
 
-fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     task::block_on(async {
         let mut res = surf::get("https://httpbin.org/get").await?;
         dbg!(res.body_string().await?);
@@ -79,7 +79,7 @@ type directly.
 ```rust
 use async_std::task;
 
-fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     task::block_on(async {
         dbg!(surf::get("https://httpbin.org/get").recv_string().await?);
         Ok(())
@@ -93,7 +93,7 @@ Both sending and receiving JSON is real easy too.
 use async_std::task;
 use serde::{Deserialize, Serialize};
 
-fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     #[derive(Deserialize, Serialize)]
     struct Ip {
         ip: String
@@ -118,7 +118,7 @@ And even creating streaming proxies is no trouble at all.
 ```rust
 use async_std::task;
 
-fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     task::block_on(async {
         let reader = surf::get("https://img.fyi/q6YvNqP").await?;
         let res = surf::post("https://box.rs/upload").body(reader).await?;
@@ -129,11 +129,11 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
 
 ## Installation
 
-Install OpenSSL - 
+Install OpenSSL -
 - Ubuntu - ``` sudo apt install libssl-dev ```
 - Fedora - ``` sudo dnf install openssl-devel ```
 
-Make sure your rust is up to date using: 
+Make sure your rust is up to date using:
 ``` rustup update ```
 
 With [cargo add](https://github.com/killercup/cargo-edit#Installation) installed :
--- a/examples/hello_world.rs
+++ b/examples/hello_world.rs
@@ -2,13 +2,13 @@ use async_std::task;
 
 // The need for Ok with turbofish is explained here
 // https://rust-lang.github.io/async-book/07_workarounds/03_err_in_async_blocks.html
-fn main() -> Result<(), surf::Error> {
+fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     femme::start(log::LevelFilter::Info)?;
 
     task::block_on(async {
         let uri = "https://httpbin.org/get";
         let string: String = surf::get(uri).recv_string().await?;
         println!("{}", string);
-        Ok::<(), surf::Error>(())
+        Ok(())
     })
 }
--- a/examples/hello_world.rs
+++ b/examples/hello_world.rs
@@ -1,8 +1,10 @@
 use async_std::task;
 
+type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
 // The need for Ok with turbofish is explained here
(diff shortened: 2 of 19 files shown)


stjepang:
What is the reason for re-exporting the error type from anyhow rather than wrapping it in a new type?

dtolnay:
Ditto -- you want this wrapped in your own type.

derekdreery:
@dtolnay do you mean like this?

derekdreery:
Since a boxed trait object is `'static` by default, I could remove all the annotations in examples, if that was wanted.

yoshuawuyts:
@derekdreery replacing the custom error types in the examples with this new error type would be v. welcome!

derekdreery:
@yoshuawuyts 
> replacing the custom error types in the examples with this new error type would be v. welcome!

I can do this, but I would argue that we don't need to. The user of this crate (e.g. the examples) can use whatever error system they like - they are not tied to using `surf::Error`.

src/error.rs
```diff
@@ -0,0 +1,59 @@
+
+use std::{fmt::{self, Display, Debug}, ops::{Deref, DerefMut}};
+
+/// A generic error type.
+pub struct Error(anyhow::Error);
+
+impl Error {
+    /// Use this when you need to implement middleware, where the error type must be `surf::Error`.
+    pub fn new<E>(error: E) -> Self
```
stjepang: Let's make this constructor private or `pub(crate)`.

src/error.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+}
+
+impl Debug for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        Debug::fmt(&self.0, f)
+    }
+}
+
+impl Deref for Error {
```
stjepang: It's not idiomatic to implement `Deref` and `DerefMut` on error types - only smart pointers should implement those traits.

stjepang:
In the examples, this is a little bit verbose:

```rust
async fn post_json() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
    // ...
}
```

We could simplify with:

```rust
type Error = Box<dyn std::error::Error + Send + Sync>;

async fn post_json() -> Result<(), Error> {
    // ...
}
```

Btw, I think it'd also be nice to have this in the crate :)
```rust
type Result<T> = std::result::Result<T, crate::Error>;
```

src/error.rs
derekdreery: I originally did this, but I found that the user needed to be able to create `surf::Errors` in the middleware examples. I'll have a look to see if this can be avoided.

derekdreery:
I've been looking at using a boxed error in middleware rather than `surf::Error`, but I run into a problem converting `Box<dyn std::error::Error + Send + Sync>` into a `anyhow::Error`. Is this possible? I would imagine not, since trait object representation is not stable (e.g. see https://github.com/rust-lang/rfcs/pull/2594 for recent work).

Otherwise the options are to make `Middleware` take an `Error` assoicated type, or to allow users to create `surf::Error`s.

Other suggestions welcome!

dtolnay:
> but I run into a problem converting `Box<dyn std::error::Error + Send + Sync>` into a `anyhow::Error`. Is this possible? I would imagine not

[`anyhow!`](https://docs.rs/anyhow/1.0.24/anyhow/macro.anyhow.html) will do that conversion.

```rust
use anyhow::{anyhow, Result};

fn f() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
    Err(Box::from("oh no!"))
}

fn main() -> Result<()> {
    match f() {
        Err(err) => Err(anyhow!(err)),
        Ok(_) => unreachable!(),
    }
}
```

derekdreery:
Cool! I've included that, so that the middleware can take any boxed error. It could still be switched to an associated type if people think that would be better. I've used an alias for boxed errors where it's hidden from the user to make the code less verbose, but where the user sees it I've always used the original type rather than an alias.

examples/hello_world.rs
```diff
@@ -1,14 +1,16 @@
 use async_std::task;
 
+type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
 // The need for Ok with turbofish is explained here
 // https://rust-lang.github.io/async-book/07_workarounds/03_err_in_async_blocks.html
-fn main() -> Result<(), surf::Exception> {
+fn main() -> Result<(), BoxError> {
```
yoshuawuyts: Could we use `surf::Error` here?

README.md
```diff
@@ -118,7 +118,7 @@ And even creating streaming proxies is no trouble at all.
 ```rust
 use async_std::task;
 
-fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
```
yoshuawuyts: Could we use `surf::Error` here?

examples/middleware.rs
```diff
@@ -4,13 +4,15 @@ use surf::middleware::{HttpClient, Middleware, Next, Request, Response};
...
+type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
 impl<C: HttpClient> Middleware<C> for Printer {
     fn handle<'a>(
         &'a self,
         req: Request,
         client: C,
         next: Next<'a, C>,
-    ) -> BoxFuture<'a, Result<Response, surf::Exception>> {
+    ) -> BoxFuture<'a, Result<Response, BoxError>> {
```
yoshuawuyts: Could we use `surf::Error` here?

examples/next_reuse.rs
```diff
@@ -5,13 +5,15 @@ use surf::middleware::{Body, HttpClient, Middleware, Next, Request, Response};
...
+type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
 impl<C: HttpClient> Middleware<C> for Doubler {
     fn handle<'a>(
         &'a self,
         req: Request,
         client: C,
         next: Next<'a, C>,
-    ) -> BoxFuture<'a, Result<Response, surf::Exception>> {
+    ) -> BoxFuture<'a, Result<Response, BoxError>> {
```
yoshuawuyts: Could we use `surf::Error` here?

src/error.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        DerefMut::deref_mut(&mut self.0)
+    }
+}
+
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        self.0.source()
+    }
+}
```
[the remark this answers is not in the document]
austinsheep: Should there be `From<Error>` impls for the log errors and the io errors in the `next_reuse` example?

derekdreery:
This is out of date so I'm gonna close.

PR #119, opened 2019-11-29
Export NativeClient

xyaman:
It could be better if we export `NativeClient`, that way we could create custom wrappers for `surf::Client`


yoshuawuyts:
Thanks for putting in this work! We've recently created a new crate: `http-client` which provides much the same functionality. https://github.com/http-rs/surf/pull/124 is a patch that helps us move towards that. Going to go ahead and close this in favor of that.

PR #124, opened 2019-12-02
Move to http-client crate

pwoolcoc:
This PR removes the `http-client` module and uses the new `http-client` crate instead

PR #125, opened 2019-12-04
Add unstable feature and export NativeClient with it

### LINKED ISSUE #96: make NativeClient public
I have a problem when I want to wrap my client.

The Problem like this: https://stackoverflow.com/questions/45253784/expected-type-parameter-found-struct.

And my code:

```
use surf::{middleware::HttpClient, Client};

pub struct NodeRpcClient2<C: HttpClient> {
    client: Client<C>,
    upstream_url: String,
}

impl<C: HttpClient> NodeRpcClient2<C> {
    pub fn new(url: String) -> Self {
        let client = Client::new();
        NodeRpcClient2 {
            client,
            upstream_url: url,
        }
    }
}

```

#69 seems to work.

XAMPPRocky:
Resolves #96 

Hello, I want to be able to wrap the surf client struct in another struct in a project I'm working on. Currently this type is not exposed because of stability guarentees. This PR adds a `unstable` feature flag and exports the `NativeClient` type with this feature enabled. This gives users a way to opt in to being able to reference the type, without requiring semver compatibility.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ hyper-client = ["hyper", "hyper-tls", "native-tls", "runtime", "runtime-raw", "r
 curl-client = ["isahc"]
 wasm-client = ["js-sys", "web-sys", "wasm-bindgen", "wasm-bindgen-futures"]
 middleware-logger = []
+unstable = []
 encoding = ["encoding_rs"]
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


yoshuawuyts:
https://github.com/http-rs/surf/pull/124 has been merged which adds the same functionality as this patch. We'll likely be putting out a 2.0 alpha soonish to resolve this. Thanks heaps for putting in this work, and apologies this has all been taking longer than should've been the case!

PR #127, opened 2019-12-06
Set Content-Length if we can

matklad:
Hi!

This PR adds tracking of body's size, which should allow the client to set `Content-Length` properly. The actual plumbing is only implemented for `isahc` client. 

This PR interacts with https://github.com/http-rs/surf/pull/124, which moves the relevant code to another repository. It's probably best to merge 124 first, and then re-apply this diff onto `surf/http-client` repo!


matklad:
closing in favor of https://github.com/http-rs/http-client/pull/15

Issue #107, opened 2019-11-20
2.0 release tracking


yoshuawuyts:
This is a tracking list of work we need to do for a 2.0 release. It's not much, and most of the API will stay the same. But still, if we have multiple changes that come together it's best to batch them (: 

## Tasks
- [x] upgrade isahc to a 0.3.0 compatible version https://github.com/http-rs/surf/issues/100
- [x] move to [`http-client`](https://github.com/http-rs/http-client/issues/2) (ref https://github.com/http-rs/surf/pull/66)
- [x]  remove generic param and use dynamic dispatch instead https://github.com/http-rs/surf/issues/69
- [x]  replace `surf::Exception` with `surf::Error` that is a re-export of `anyhow::Error` https://github.com/http-rs/surf/issues/86
- [x] use `surf::Request` and `surf::Response` in middleware #131 
- [ ] Connection pooling for Client / HttpClient https://github.com/http-rs/surf/issues/224
- [ ] Possibly use the async-h1 HttpClient by default rather than Isahc https://github.com/http-rs/surf/issues/217
- ~~[ ] wasm compatibility #210~~

Fishrock123:
Updated the list slightly, but I think this is better tracked in the milestone. See https://github.com/http-rs/surf/milestone/1

PR #128, opened 2019-12-11
replace http_client with http-client crate

zkat:
This should help with #107 a bit.

I do need some help getting this fully working, though. I'm not sure I'm adding the dependency correctly, for one, and for two, the `AsyncReadExt` trait import I put in `response.rs` doesn't seem to be working for me:

```
➜ cargo check                                           
    Checking surf v1.0.3 (/Users/zkat/Documents/code/personal/rust/surf)
error[E0599]: no method named `read_to_end` found for type `&mut http_client::Body` in the current scope
   --> src/response.rs:139:34
    |
139 |         self.response.body_mut().read_to_end(&mut buf).await?;
    |                                  ^^^^^^^^^^^ method not found in `&mut http_client::Body`
    |
    = note: the method `read_to_end` exists but the following trait bounds were not satisfied:
            `&mut http_client::Body : futures_util::io::AsyncReadExt`
            `http_client::Body : futures_util::io::AsyncReadExt`
    = help: items from traits can only be used if the trait is in scope
    = note: the following trait is implemented but not in scope, perhaps add a `use` for it:
            `use futures_util::io::AsyncReadExt;`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: could not compile `surf`.
```


pwoolcoc:
I already have a PR open for this here: https://github.com/http-rs/surf/pull/124

yours isn't compiling because we need a new version of `http-client` that uses the same version of `futures` as surf does.

zkat:
Omg sorry I missed that for some reason.

PR #133, opened 2019-12-24
Convert `Exception` to `Error` cont.

### LINKED ISSUE #86: surf::Exception does not implement std::error::Error

austinsheep:
Continuation of #113 to resolve #86.

In my commit I changed all instances of `BoxError` with `Error`.  I also added `impl From`'s for `log::SetLoggerError` and `std::io::Error`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ serde = "1.0.97"
 serde_json = "1.0.40"
 serde_urlencoded = "0.6.1"
 url = "2.0.0"
+anyhow = "1.0.23"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 # encoding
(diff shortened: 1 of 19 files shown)


Fishrock123:
We moved to [`http_types::Error`](https://docs.rs/http-types/2.3.0/http_types/struct.Error.html) in https://github.com/http-rs/surf/commit/c8366fb2b9d7d59af8e9cc7a5f9d2173d459e7ec.

Thanks anyways!

Issue #97, opened 2019-11-09
`<Type> cannot be shared between threads safely` during middleware definition


zkat:
I'm working on a caching middleware for Surf, and got a pretty inscrutable error. See https://gist.github.com/zkat/f9e71856bdca23eb2b665ece262db091

From a conversation in Discord, it sounds like this is because Body doesn't have a Sync bound?

yoshuawuyts:
@zkat yeah that def seems to be the case. We should add that bound. Unsure if that's a breaking change, but we should patch this for sure.

Going to put this under the 2.0 banner just to be safe.

Fishrock123:
Was fixed in #138

PR #138, opened 2020-01-10
Fix body bounds

yoshuawuyts:
Closes https://github.com/http-rs/surf/issues/97. Thanks!


goto-bus-stop:
Looks like this was done when adding the h1 backend :D

PR #141, opened 2020-01-30
Remove default dependency on native-client

tamird:
As a side effect, this allows surf to compile without default features.

NOTE: the above is not currently true because doctests depend on
native-client and middleware-logger. Any suggestions on how to plumb a
feature requirement into doctests?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,32 +12,51 @@ readme = "README.md"
 edition = "2018"
 
 [features]
-default = ["native-client", "middleware-logger", "encoding"]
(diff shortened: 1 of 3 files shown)


yoshuawuyts:
@tamird Hi, thanks for opening this issue. I'm not sure if this is the right direction for surf? I don't feel confident merging this right now, as we're a bit in flux, especially in the light of #146. Apologies, and thanks again!

tamird:
@yoshuawuyts the current state of surf is that it doesn't compile with `default-features = false`. Isn't that worthy of fixing?

PR #145, opened 2020-02-27
reexport NativeClient

clouds56:


andrewbanchich:
It looks like this is [no longer necessary](https://docs.rs/surf/2.0.0/surf/struct.Client.html) due to changes in 2.0.

Fishrock123:
Yeah, see https://crates.io/crates/http-client

PR #154, opened 2020-03-23
Add optional simd-json support

Licenser:
This adds the `simd-json` as a feature flag to use SIMD accelerated JSON parsing.

This might warrant a note in the README about cpu features required (AVX2 or SSE4.1) to use the flag but I didn't find a section in the README on features and wasn't sure if you want one added so I'll just open this for suggestions :).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ serde_json = "1.0.40"
 serde_urlencoded = "0.6.1"
 url = "2.0.0"
 http-client = { version = "1.1.0", features = ["native_client"] }
-simd-json = { version = "0.3.2", optional = true }
+simd-json = { version = "0.3.7", optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 # encoding
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ serde_json = "1.0.40"
 serde_urlencoded = "0.6.1"
 url = "2.0.0"
 http-client = { version = "1.1.0", features = ["native_client"] }
+simd-json = { version = "0.3.2", optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 # encoding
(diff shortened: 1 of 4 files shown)


skade:
This is a surprisingly simple addition and looks good from my side.

Licenser:
I'll add docs to the lib! did miss that sorry

Licenser:
I've added the same documentation to https://github.com/http-rs/tide/pull/423

Licenser:
I recently added support for `Serialize` in simd-json and updated the PR to reflect that.

dignifiedquire:
Any reason not to enable this by default?

dignifiedquire:
docs are still missing

CryZe:
> Any reason not to enable this by default?
Yes, it doesn't compile if you don't have SSE 4.2 or AVX 2 on your platform, unless you activate a feature to allow for a non-SIMD fallback (neither of which you have with a default compilation on any platform). Also the crate requires nightly for ARM targets.

dignifiedquire:
@CryZe I see, thanks

Licenser:
We put measures in place in `simd-json` to not allow compilation when neither  SSE4.2 or AVX2 are available to prevent people from having an unexpectedly slow experience so making it the default would leave people in the sad place that out of the box surf might not compile and throw errors and I'd like to avoid that :D,

re: docs - sorry I am switching computers too much these days I wrote them and must have then pushed from somewhere else 🤦 I'll add them!

Licenser:
One thing I'm a bit unsure about ergonomics wise is if `serde_json` should be optional and a default feature flag.

I see two sides to this (here and in http-rs/tide#423), on the one side, if the `simd-json` feature is used, right now it pulls in an unused dependency (even so it won't get compiled), that feels a bit redundant but not too bad.

On the other side making it a feature might break users build if they are using `--no-default-features` which I'm quite hesitant to risk. It has, however, the upside that users that do not need any build-in JSON could just skip both?

I personally tend to stick to the first option  (the one the PR does it right now) as it is kinder to users even so an unused dependency is included or wait to a major release that includes breaking changes anyway to feature-out JSON completely. 

Happy to change it if the preference here is a different one :)

PR #156, opened 2020-03-26
Switch to a concrete Error type

belak:
Please note that this is specifically a PoC to replace usage of `Box<dyn std::error::Error + Send + Sync + 'static>` with a concrete `surf::Error` type.

There are a few things yet to figure out:

* Documentation for `surf::{Result, Error}`
* Whether or not the Error variants should be exposed to the end user or if they should be hidden. I went with hiding it because it's similar to `anyhow::Error`.
* When to return a `surf::Error` vs alternatives (like `serde_json::Error`) or if `surf::Error` should always be used
* How errors should be displayed. Should they be wrapped or displayed as-is?

The main `Error` implementation is in `error.rs` and is also exported at the root of the crate.

This is an alternative to simply re-exporting `anyhow::Error` and would be another method of fixing #86.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,43 @@
+use std::result::Result as StdResult;
+
+use thiserror::Error as ThisError;
+
+#[allow(missing_docs)]
+pub type Result<T> = StdResult<T, Error>;
+
+#[allow(missing_docs)]
+#[derive(ThisError, Debug)]
+#[error(transparent)]
+pub struct Error(#[from] InternalError);
+
+macro_rules! impl_from {
+    ($bound:ty) => {
+        impl From<$bound> for Error {
+            fn from(err: $bound) -> Error {
+                InternalError::from(err).into()
+            }
+        }
+    };
+}
+
+impl_from!(log::kv::Error);
+impl_from!(serde_urlencoded::ser::Error);
+impl_from!(serde_urlencoded::de::Error);
+impl_from!(std::io::Error);
+impl_from!(serde_json::Error);
+
+#[derive(ThisError, Debug)]
+pub(crate) enum InternalError {
+    #[error(transparent)]
+    LogError(#[from] log::kv::Error),
+    #[error(transparent)]
+    UrlEncodeError(#[from] serde_urlencoded::ser::Error),
+    #[error(transparent)]
+    UrlDecodeError(#[from] serde_urlencoded::de::Error),
+    #[error(transparent)]
+    IoError(#[from] std::io::Error),
+    #[error(transparent)]
+    JsonError(#[from] serde_json::Error),
+    #[error("{0}")]
+    HttpError(String),
+}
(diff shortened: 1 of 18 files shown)


zkat:
hey @Fishrock123 @yoshuawuyts can we take this one? I think this would solve the error issue nicely.

@belak I notice you have a lot of conflicts. Are you ok with fixing them, or would you rather someone else do it, at this point?

Fishrock123:
At a glance this sounds right to me, that’s what Tide does.

belak:
I'd be happy to work on this, but it might take a few days

src/error.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use thiserror::Error as ThisError;
+
+#[allow(missing_docs)]
+pub type Result<T> = StdResult<T, Error>;
+
+#[allow(missing_docs)]
+#[derive(ThisError, Debug)]
+#[error(transparent)]
+pub struct Error(#[from] InternalError);
```
Fishrock123, line 11: This probably doesn't need to exist. Ideally this would just using `http_types::Error` instead:  https://docs.rs/http-types/2.3.0/http_types/struct.Error.html

(It's implemented like this: https://github.com/http-rs/http-types/blob/master/src/error.rs)

src/error.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+impl_from!(log::kv::Error);
+impl_from!(serde_urlencoded::ser::Error);
+impl_from!(serde_urlencoded::de::Error);
+impl_from!(std::io::Error);
+impl_from!(serde_json::Error);
+
+#[derive(ThisError, Debug)]
+pub(crate) enum InternalError {
```
Fishrock123, line 30: Also with `http_types::Error` this shouldn't be necessary, because it wraps the errors and allows you to check [downcasts](https://docs.rs/http-types/2.3.0/http_types/struct.Error.html#method.downcast_ref), like so:
```rust
match surf::get(url).await {
    Ok() -> (),
    Err(error) -> {
        if let Some(&url_err) = error.downcast_ref::<UrlError>() {
            // ...
        } else if let Some(&json_err) = error.downcast_ref::<JsonError>() {
            // ...
        }
    }
}
```

src/error.rs
```diff
@@ -0,0 +1,43 @@
+use std::result::Result as StdResult;
+
+use thiserror::Error as ThisError;
+
+#[allow(missing_docs)]
+pub type Result<T> = StdResult<T, Error>;
```
Fishrock123, line 6: This [already exists](https://github.com/http-rs/surf/blob/3a8a25b40c5f927fb0e425547c5990bcc577fc73/src/lib.rs#L100-L101) from the recent-ish https://github.com/http-rs/surf/commit/ac215cbaa1f612364aff7e1bf5ec282c862f8c90

Fishrock123:
Closing since as noted above, the better change would have been a switch to `http_types::Error`, which was already done in April in https://github.com/http-rs/surf/commit/c8366fb2b9d7d59af8e9cc7a5f9d2173d459e7ec!

PR #157, opened 2020-03-30
Remove unused dependency feature

oli-cosmian:
This feature is unused in `surf` and causes havoc with our dependency on `paperclip` (namely `paperclip` doesn't compile due to some type inference problems)


oli-cosmian:
Hmm... this may have been a bit overzealous by me. The reason it still works is because `async-std` has a default feature activating this feature on `log`. I'll create a real fix.

goto-bus-stop:
Oh! I'll revert it for now then in a108576. Thanks!

PR #159, opened 2020-04-04
Update isahc to 0.9 and http to 0.2.1

### LINKED ISSUE #153: Update isahc depedency from 0.8 to 0.9
[Isahc 0.9 has been released with some breaking changes](https://github.com/sagebind/isahc/releases/tag/0.9.0), most notable with dependency on http updated from 0.1 to 0.2.

It would be cool if surf followed suit so that the world can move away from the legacy packages.

mythmon:
My approach here was fairly naive. I read the changelog for the two libraries, and then updated the files in Cargo.toml and made changes until it compiled. It's quite possible I've missed something.

This patch depends on [my PR to http-client](https://github.com/http-rs/http-client/pull/24) merging and being released. For now I've set the http-client dependency to point to my git branch, but of course that would need to change before this merges.

Fixes #153

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ encoding = ["encoding_rs"]
 
 [dependencies]
 futures = { version = "0.3.1", features = ["compat", "io-compat"] }
-http = "0.1.17"
+http = "0.2.1"
 log = { version = "0.4.7", features = ["kv_unstable"] }
 mime = "0.3.13"
 mime_guess = "2.0.3"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -22,22 +22,22 @@ encoding = ["encoding_rs"]
...
+http = "0.2.1"
 log = { version = "0.4.7", features = ["kv_unstable"] }
 mime = "0.3.13"
 mime_guess = "2.0.3"
 serde = "1.0.97"
 serde_json = "1.0.40"
 serde_urlencoded = "0.6.1"
 url = "2.0.0"
-http-client = { version = "1.1.0", features = ["native_client"] }
+http-client = { git = "https://github.com/mythmon/http-client", branch = "isahc-0.9", features = ["native_client"] }
```
mythmon, line 33: Notably, this line will need to be updated to something like this, once my PR to http-client merges and is released.

```toml
http-client = { version = "2.0.0", features = ["native_client"] }
```

Licenser:
<3 http 0.2 would be lovely

goto-bus-stop:
Thanks for the PR! This was mostly done in https://github.com/http-rs/http-client/pull/22 (released in http-client 2.0.0) and https://github.com/http-rs/surf/pull/146, though we still have an unused dependency on isahc 0.8 in surf itself which is being removed in https://github.com/http-rs/surf/pull/164.

PR #160, opened 2020-04-10
Cookies

### LINKED ISSUE #19: Middleware: cookie jar
We should write a middleware that can persist cookies between runs.

basdebue:
This pull request adds support for managing cookies and closes #19.

Progress:
- [x] Add cookie-related methods to `Request` and `Response`
- [ ] Implement 'cookie jar' middleware

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ middleware-logger = []
 encoding = ["encoding_rs"]
 
 [dependencies]
+cookie = "0.13.3"
 futures = { version = "0.3.1", features = ["compat", "io-compat"] }
 http = "0.1.17"
 log = { version = "0.4.7", features = ["kv_unstable"] }
--- a/src/request.rs
+++ b/src/request.rs
@@ -1,3 +1,4 @@
+use cookie::Cookie;
 use futures::future::BoxFuture;
 use futures::prelude::*;
 use http::Method;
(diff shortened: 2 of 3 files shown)


src/request.rs
```diff
@@ -32,6 +33,8 @@ pub struct Request<C: HttpClient + Debug + Unpin + Send + Sync> {
     client: Option<C>,
     /// Holds the state of the request.
     req: Option<http_client::Request>,
+    /// Holds the cookies added to the request.
+    cookies: Option<Vec<Cookie<'static>>>,
```
basdebue, line 37: I have opted for a `'static` lifetime here as to not introduce lifetime bounds to the struct. This does however mean that we potentially have to allocate twice in the `set_cookie` method to obtain the required `String`s.

src/response.rs
```diff
@@ -16,13 +17,20 @@ use http_client;
...
 pub struct Response {
+    cookies: Vec<Cookie<'static>>,
     response: http_client::Response,
 }
 
 impl Response {
     /// Create a new instance.
     pub(crate) fn new(response: http_client::Response) -> Self {
-        Self { response }
+        let cookies = response
```
basdebue, line 27: Separating the cookies from the rest of the response here means we can hand out `&str`s in the cookie-related methods. Unfortunately, this isn't zero-cost as it will allocate once per received cookie, even if you don't intend on using them at all.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ middleware-logger = []
 encoding = ["encoding_rs"]
 
 [dependencies]
+cookie = "0.13.3"
```
basdebue, line 24: Should this dependency be feature-gated?

basdebue:
Because `surf`'s internals have changed quite a bit since I first submitted this PR, I've decided to start over instead of doing the complex rebasing work to continue with this branch. I'll be submitting a new PR soon!

PR #164, opened 2020-04-22
Clean up dependencies

goto-bus-stop:
- Remove dependencies that had been moved into http-client.
- Use the `http-client/*_client` features for curl/wasm/h1.
- Remove mentions of hyper entirely as it is not currently implemented.
- Test both `native-client` and `h1-client` on CI.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,8 @@ readme = "README.md"
 edition = "2018"
 
 [features]
+# when the default feature set is updated, verify that the `--features` flags in
+# `.github/workflows/ci.yaml` are updated accordingly
 default = ["native-client", "middleware-logger", "encoding"]
 h1-client = ["wasm-client", "http-client/h1_client"]
 native-client = ["curl-client", "wasm-client", "http-client/native_client"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ default = ["native-client", "middleware-logger", "encoding"]
 h1-client = ["wasm-client", "http-client/h1_client"]
 native-client = ["curl-client", "wasm-client", "http-client/native_client"]
 curl-client = ["http-client/curl_client"]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -12,14 +12,16 @@ readme = "README.md"
...
 native-client = ["curl-client", "wasm-client", "http-client/native_client"]
-hyper-client = ["hyper", "hyper-tls", "native-tls", "runtime", "runtime-raw", "runtime-tokio" ]
-curl-client = ["isahc"]
-wasm-client = ["js-sys", "web-sys", "wasm-bindgen", "wasm-bindgen-futures"]
+curl-client = ["http-client/curl_client"]
+wasm-client = ["web-sys", "http-client/wasm_client"]
 middleware-logger = []
-encoding = ["encoding_rs"]
+# requires web-sys for TextDecoder on wasm
+encoding = ["encoding_rs", "web-sys"]
```
yoshuawuyts, line 24: Won't this also pull in `web-sys` on non-wasm targets? I believe if `wasm-client` has been enabled then `web-sys` is already available.

Cargo.toml
```diff
@@ -33,53 +35,25 @@ serde_urlencoded = "0.6.1"
...
-    "ReferrerPolicy",
-    "Request",
-    "RequestCache",
-    "RequestCredentials",
-    "RequestInit",
-    "RequestMode",
-    "RequestRedirect",
-    "Response",
     "TextDecoder",
-    "Window",
```
yoshuawuyts, line 78: Oh wow; are these features no longer needed? That's huge!

goto-bus-stop: They're all specified in http-client :D

Cargo.toml
goto-bus-stop: web-sys is still listed as a dependency for wasm targets only—appears to work as expected this way. actually web-sys is no longer directly used by surf for the wasm-client feature, so we can remove it from there

PR #171, opened 2020-05-04
feat: Add Response#header_names()

### LINKED ISSUE #166: No way to iterate over response headers
There's currently no way to iterate over surf headers. I'm not sure how often this is used in practice, but example use cases might be:

- a command line tool built around surf that prints out headers in a verbose mode
- a proxy library that performs some sort of iterator transformation, such as "pass through some but not all of the headers from my upstream, depending on some predicate"
- a third party tool that wants to log or serialize information about a surf::Response

Since none of these are common needs, this also would be addressed by making surf::Response `Into<http_types::Response>` (and/or `Deref<Target=http_types::Response>`, but the case for that is less compelling)

chrisdickinson:
This allows users to enumerate response headers – we're just passing the underlying `header_names` method through.

Fixes #166.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -87,7 +87,7 @@ impl Response {
     /// assert!(names.contains(&&"Content-Length".parse().unwrap()));
     /// # Ok(()) }
     /// ```
-    pub fn header_names(&'_ self) -> Names<'_> {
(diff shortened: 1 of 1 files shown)


Fishrock123:
This functionality has been merged via https://github.com/http-rs/surf/pull/191

PR #185, opened 2020-05-31
Fix wasm compilation

Perseus101:
Blocked by https://github.com/http-rs/http-client/pull/28


Perseus101:
This is fixed with the release of http-client v5.0

PR #187, opened 2020-06-03
Fix set header

### LINKED ISSUE #183: Runtime panic with alpha.3 when setting header
I'm getting a panic at runtime when attempting to set a header on a `Request`. I could be misusing the library so please let me know if that's the case.

It looks like the error might be coming from [http_types::Request::insert_header](https://docs.rs/http-types/2.1.0/http_types/struct.Request.html#method.insert_header). 

I've tried `cargo clean`/`rm Cargo.lock`/`cargo build` to no avail.

The following code panics at runtime in alpha.3:

```rust
let response = surf::Request::new(method, url)
    .set_header("Content-Type", "application/json")
    .recv_json().await?;
```

The pertinent information from the backtrace appears to be this:

```
surf::request::Request<C>::set_header
    /home/james/.cargo/registry/src/github.com-1ecc6299db9ec823/surf-2.0.0-alpha.3/src/request.rs:200
```

If I comment out the `set_header` line, the request goes through (although the `recv_json` breaks now because the server doesn't respond properly).

The following code works in alpha.2:

```rust
let response = surf::Request::new(method, url)
    .set_header("Content-Type".parse().unwrap(), "application/json")
    .recv_json().await?;
```

Licenser:
This fixes #183 - a panic set-header in 0.3. The issue was that `set_header` calls `unwrap()` on the option returned by `insert_header`, which only returns a `Some` if a header already existed.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -197,11 +197,7 @@ impl<C: HttpClient> Request<C> {
     /// # Ok(()) }
     /// ```
     pub fn set_header(mut self, key: impl Into<HeaderName>, value: impl ToHeaderValues) -> Self {
-        self.req
-            .as_mut()
-            .unwrap()
-            .insert_header(key, value)
-            .unwrap();
+        self.req.as_mut().unwrap().insert_header(key, value);
         self
     }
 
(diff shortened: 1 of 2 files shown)


Licenser:
@yoshuawuyts any chance to fast track this fix since the underlying bug makes `alpha.3` unusable when headers are involved?

yoshuawuyts:
I'll try and get a release for this in on Friday.

Licenser:
Thank you! <3

yoshuawuyts:
published!

Licenser:
Thank you so much for the quick handling!!

PR #182, opened 2020-05-28
feat: add header-related methods to Response

SirWindfield:
This PR adds various methods for working with the received headers from a response object.
An iterator over the names of each header field has been added in PR #171.

Closes #149.

PR #174, opened 2020-05-08
Add an `add_header` method to `Request`

pwoolcoc:
This corresponds to http-types `append_header` method, and will assist
in correctly converting `http_types::Request` objects to `surf::Request`
objects.

PR #191, opened 2020-06-29
Request, Response: method parity with http-types/tide

Fishrock123:
Adds many methods on http_types request/response that Tide exposes.

Supercedes the following PRs:
- #182
- #174
- #171

Also changes the "mime" / "content_type" apis to use `http_types::mime` like Tide, which is a breaking change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -56,7 +56,7 @@
 //! # #[async_std::main]
 //! # async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
 //! let req = surf::get("https://img.fyi/q6YvNqP").await?;
-//! let body = surf::http_types::Body::from_reader(req, None);
+//! let body = surf::http::Body::from_reader(req, None);
 //! let res = surf::post("https://box.rs/upload").set_body(body).await?;
 //! # Ok(()) }
 //! ```
@@ -83,9 +83,9 @@ mod response;
 
 pub mod middleware;
 
-pub use http_types;
-pub use http_types::mime;
-pub use http_types::{Error, Result};
+#[doc(inline)]
+pub use http_types::{self as http, Body, Error, Status, StatusCode};
+
 pub use url;
 
 pub use client::Client;
@@ -96,3 +96,6 @@ pub use response::{DecodeError, Response};
 mod one_off;
 #[cfg(any(feature = "native-client", feature = "h1-client"))]
 pub use one_off::{connect, delete, get, head, options, patch, post, put, trace};
+
+/// A specialized Result type for Surf.
+pub type Result<T = Response> = std::result::Result<T, Error>;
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -55,8 +55,9 @@
 //! ```no_run
 //! # #[async_std::main]
 //! # async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
-//! let reader = surf::get("https://img.fyi/q6YvNqP").await?;
-//! let res = surf::post("https://box.rs/upload").body(reader).await?;
+//! let req = surf::get("https://img.fyi/q6YvNqP").await?;
+//! let body = surf::http_types::Body::from_reader(req, None);
```
yoshuawuyts, line 59: we should re-export `http-types` as `surf::http`

src/lib.rs
```diff
@@ -83,8 +84,8 @@ mod response;
 pub mod middleware;
 
 pub use http_types;
+pub use http_types::mime;
```
yoshuawuyts, line 87: We're not re-exporting this from tide either -- though perhaps we should? I'm curious about the reasoning here.

src/response.rs
```diff
@@ -1,31 +1,30 @@
...
-        response: http_client::Response,
+        res: http_client::Response,
     }
 }
 
 impl Response {
     /// Create a new instance.
     pub(crate) fn new(response: http_client::Response) -> Self {
-        Self { response }
+        Self { res: response }
```
yoshuawuyts, line 27: Renaming `response` to `res` may be the simpler change here.

src/response.rs
```diff
@@ -155,14 +233,14 @@ impl Response {
     /// let string: String = res.body_string().await?;
     /// # Ok(()) }
     /// ```
-    pub async fn body_string(&mut self) -> Result<String, Error> {
+    pub async fn body_string(&mut self) -> http_types::Result<String> {
```
yoshuawuyts, line 236: Is `surf::Result` not available? That should probably be a re-export of `http_types::Result`.

src/response.rs
```diff
@@ -155,14 +233,14 @@ impl Response {
...
-        let mime = self.mime();
+        let mime = self.content_type();
         let claimed_encoding = mime
             .as_ref()
-            .and_then(|mime| mime.get_param("charset"))
-            .map(|name| name.as_str());
-        decode_body(bytes, claimed_encoding)
+            .and_then(|mime| mime.param("charset"))
+            .map(|name| name.to_string());
+        decode_body(bytes, claimed_encoding.as_deref())
```
yoshuawuyts, line 243: Neat; excited you managed to translate this. We should probably consider adding this to `http-types` eventually (believe it's the longest-open issue right now).

src/response.rs
Fishrock123: Tide's is not a re-export, I'm going to change it to be like Tide: https://github.com/http-rs/tide/blob/e994b2ebbaa5a1ebe6849bff720adff89dae1fb5/src/lib.rs#L286-L287

src/response.rs
Fishrock123: It's a tad tricky there due to the absence of headers information from `Body`. maybe on `http_types::request` though.

Fishrock123:
- missing `ext`

Issue #192, opened 2020-07-08
Rework `Request` (and related)


Fishrock123:
This is essentially a summary of the major improvements to Surf I've been trying to make. there is a lot of overlap with disparate existing issues, which I've linked where appropriate.

--------

In order to improve middleware for Surf (https://github.com/http-rs/surf/issues/131), and bring it more in-line with Tide, the following changes should be made:
- Middleware should operate on  `surf::Request`, `surf::Response` rather than `http_types::Request`, etc.
- `surf::Request` should be changed to be a more minimalized wrapper around `http_types::Request` (similar to `Response` and Tide.)

And also for code hygiene, `Request` should probably not be all of these at once (https://github.com/http-rs/surf/issues/152):
- the sent http structure
- the builder
- the holder of reusable resources (Client, Middleware) (https://github.com/http-rs/surf/issues/126)
- the consumable `Future`

Additionally related, `Client` should not be generic (https://github.com/http-rs/surf/issues/69), ([comment on 105](https://github.com/http-rs/surf/issues/150#issuecomment-601579021)), and Middleware _may_ want to be accessible _from the present Middleware stack_ for complete retry/redirect purposes (https://github.com/http-rs/surf/issues/169), (https://github.com/http-rs/surf/issues/18).

**Note**: Making `Client` 1:1 with an actualized `Request` helps this _but doesn't work for `keep-alive` or http2/3+_.

------

There is a clean but slightly less ergonomic way to solve most of the API issues here:
- Have `surf::Request` be like `tide::Request`, only hold `http_types::Request`.
- Do not internally hold a `surf::Request`
- Have `Client::send(req: impl Into<surf::Request>) -> BoxedFuture` rather than making anything be directly await-able.
    - (Since that requires internal storage and causes the awaited structure to be consumed.)
- Have Middleware be stored on the `Client`. (https://github.com/http-rs/surf/issues/126)
- Have a separate `RequestBuilder` similar to `tide::ResponseBuilder`.
- Have `RequestBuilder::send()` be a short-cut for "build and make a new `Client` and call `.send(self)` on it".
- Have `surf::method()` return a `RequestBuilder`.

```rust
let res = surf::post(url)
    .header(h, v)
    .body(b)
    .send().await?;
```

With middleware:

```rust
let req = surf::post(url)
    .header(h, v)
    .body(b);
let client = surf::client()
    .middleware(m);
let res = client.send(req).await?;
```

[...]

Issue #131, opened 2019-12-12
Pass `surf::Request` and `surf::Response` through middleware


yoshuawuyts:
Rather than passing `http::Request` and `http::Response` through the middleware, we should pass `surf::Request` and `surf::Response` through. This should significantly reduce the API surface we expose, and make it so folks no longer have to learn two different sets of APIs to write against.

calvinbrown085:
@yoshuawuyts Hi! Would this be hard for a newcomer to pick up? Is there a place I could start looking? I'm pretty familiar with Rust syntax and libraries, but just started looking at surf :)

goto-bus-stop:
@calvinbrown085 I think a newcomer could get pretty far with this at least! I would start by changing the signature of the `Middleware#run` trait method to use the `surf::Request` type, and letting the compiler errors tell me which other places to update.

I would expect the trickiest part to be in the [`Request#poll` implementation](https://github.com/http-rs/surf/blob/8368e88f6545e950b06e36e874c098f800a2ca47/src/request.rs#L572). That method takes ownership of some values out of the `Request` instance, _before_ running the middleware functions. It _may_ be necessary to change the ordering there (not sure, you would probably find out quick enough though...), and I don't know what lifetime issues that may cause. If you get stuck you can always ask for help in the [async-std discord](https://discord.gg/JvZeVNe).

Fishrock123:
Unfortunately this isn't that easy to do without also moving `surf::Request` to no longer consume itself in `poll`, i.e. no longer be the holder of `middleware` or the `Future`, which causes a cascade of other necessary changes. See #192.

Fishrock123:
PR at https://github.com/http-rs/surf/pull/194

Fishrock123:
Completed via #194.

PR #194, opened 2020-07-09
src: overhaul Client, rework Request and Middleware, add builder

Fishrock123:
A practical implementation of https://github.com/http-rs/surf/issues/192

Please see that issue for full background on this PR.

This includes the following changes:
- `surf::{method}()` now returns `RequestBuilder`.
- Adds `RequestBuilder`, which has `Into<Request>`.
- Changes `Middleware` to roughly `(surf::Request, surf::Client, surf::Next) -> Result<surf::Response>`.
- Changes `Endpoint` to roughly `(surf::Request, surf::Client) -> Result<surf::Response>`
- Adds `Client::send(impl Into<Request>)`
- Adds `Client::recv_{type}()`
- Adds `Client::middleware()`
- Adds `surf::client()`
- Removes `Request::middleware()` _(moved to `Client`)_
- Removed `impl Future for Request` & related _(moved to `RequestBuilder` or `Client::send()`)_
- ~~Removes `Client::{method}()` _(no longer makes much sense)_~~ Re added in 8977cba49059e1762b191855ac9fba4bb845a0c4 because people want it.
- (and more lesser changes)

**Note**: The `client: Client` passed to `Middleware` uses the same underlying `HttpClient` but does not preserve the `Middleware` stack. I do not know if it is desirable to keep it, and much less if it is possible to make such a thing work with the borrow checker.

This is absolutely semver-major.

**Edit** to be clear:
The API for one-offs looks almost identical:
```rust
let res = surf::post(url)
    .header(h, v)
    .body(b)
    .await?;
```

Re-use of a `Client` or use of `Middleware` now looks like this:
```rust
let client = surf::client()
    .middleware(m);

let req = surf::post(url)
    .header(h, v)
    .body(b);
let res = client.send(req).await?;
```

_~~Note: This presently causes stack overflows on the tests. I'm not sure why but I suspect it has to do with the middleware changes.~~ It was an accidental recursive `Client::send`._

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,8 +1,9 @@
 use std::fmt;
 use std::sync::Arc;
 
+use crate::http::Method;
 use crate::middleware::{Middleware, Next};
-use crate::{HttpClient, Request, Response, Result};
+use crate::{HttpClient, Request, RequestBuilder, Response, Result};
 
 use futures_util::future::BoxFuture;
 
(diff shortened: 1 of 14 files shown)


Fishrock123:
I have made a possible additional patch at https://github.com/Fishrock123/surf/pull/1 that re-adds `surf::{method}().middleware()` in the form of `RequestBuilder::middleware()`, and an `Option<Vec<Arc<dyn Middleware>>>` on `RequestBuilder`.

src/client.rs
```diff
@@ -1,32 +1,35 @@
...
 /// ```no_run
 /// # #[async_std::main]
 /// # async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
 /// let client = surf::Client::new();
-/// let req1 = client.get("https://httpbin.org/get").recv_string();
-/// let req2 = client.get("https://httpbin.org/get").recv_string();
-/// let (str1, str2) = futures::future::try_join(req1, req2).await?;
+/// let res1 = client.recv_string(surf::get("https://httpbin.org/get"));
+/// let res2 = client.recv_string(surf::get("https://httpbin.org/get"));
+/// let (str1, str2) = futures::future::try_join(res1, res2).await?;
```
dignifiedquire, line 25: I think it would be really nice if we could replicate the `surf::*` top level methods on the client, such that this example could be written as

```rust
let res1 = client.get("https://httpbin.org/get").recv_string();
let res2 = client.get("https://httpbin.org/get").recv_string();
```

src/request_builder.rs
```diff
@@ -0,0 +1,252 @@
...
+        Ok(res.body_form::<T>().await?)
+    }
+
+    /// Return the constructed `Request`.
+    pub fn build(self) -> Request {
+        self.req.unwrap()
+    }
+
+    /// Create a `Client` and send the constructed `Request` from it.
+    pub fn send(self) -> BoxFuture<'static, Result<Response>> {
```
dignifiedquire, line 219: Having either `send_with_client(self, client: &Client) -> BoxFuture<'static, Result<Response>>` or `with_client(self, &Client) -> Self` (storing the client and using it for `send`) would make the client reuse much nicer.

src/client.rs
Fishrock123: This unfortunately doesn't make much sense...? What would `client.get()` return? A single `Request` stored on a reusable `Client` surely isn't it. A `RequestBuilder`? That would mean having it store an `Option<Client>` too...

src/request_builder.rs
Fishrock123:
```rust
let client = surf::client();
let res1 = client.send(surf::post(url).header().body()).await?;
let res2 = client.send(surf::post(url).header().body()).await?;
```

```rust
let client = surf::client();
let res1 = surf::post(url).header().body().with_client(&client).await?;
let res1 = surf::post(url).header().body().with_client(&client).await?;
```

I dunno, not only is the bottom longer but it's also weirder, injecting a reusable struct rather than having the reusable struct do the thing...? Doesn't feel right to me...

Fishrock123:
I would like to make it clear that this _is required_ in order to have middleware use `surf::Request` and `surf::Response` rather than the `http_types` structs (https://github.com/http-rs/surf/issues/131).

Maybe the builder can be extended more to have more existing api parity or whatever, but the `Request` struct itself still need to change.

Fishrock123:
@yoshuawuyts please take a look at this

Fishrock123:
I've made a second additional patch to https://github.com/Fishrock123/surf/pull/1 which re-adds support for per-request middleware. I don't fancy the patch and think it is probably not what we should encourage users to do, ergo a mistake, but if people like it, then... 🤷

goto-bus-stop:
I agree on per-request middleware—I think that should exclusively be a property of the client and if you want a different middleware stack you should use a different client instance.

Fishrock123:
I have added 8977cba49059e1762b191855ac9fba4bb845a0c4.

It adds `client::{method}()` as a different way to do `client.send(req)`. I still don't like it but it seems to be the acceptable compromise that will allow the rest of this to move forward.

@yoshuawuyts @jbr

src/client.rs
Fishrock123: I still disagree with it but this has been done in 8977cba49059e1762b191855ac9fba4bb845a0c4, @dignifiedquire please take a look again if you are able to.

PR #195, opened 2020-07-10
middleware: add Redirect

Fishrock123:
Should work but isn't ideal. See code comments for details.

Implements https://github.com/http-rs/surf/issues/18

### DIFF
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -55,6 +55,9 @@ use std::sync::Arc;
 pub use http_client::{Body, HttpClient, Request, Response};
 
 pub mod logger;
+mod redirect;
+
+pub use redirect::Redirect;
 
 use futures::future::BoxFuture;
 use http_types::Error;
--- a/src/middleware/redirect/mod.rs
+++ b/src/middleware/redirect/mod.rs
@@ -1,0 +1,128 @@
+//! HTTP Redirect middleware.
+//!
+//! # Examples
+//!
+//! ```no_run
+//! # #[async_std::main]
+//! # async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+//! let mut res = surf::get("https://httpbin.org/redirect/2")
+//!     .middleware(surf::middleware::Redirect::default())
+//!     .await?;
+//! dbg!(res.body_string().await?);
+//! # Ok(()) }
+//! ```
+
+use std::sync::Arc;
+
+use crate::http::{headers, StatusCode};
+use crate::middleware::{HttpClient, Middleware, Next, Request, Response};
+use crate::url::Url;
+use crate::Result;
+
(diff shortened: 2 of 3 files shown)


src/middleware/mod.rs
```diff
@@ -55,6 +55,7 @@ use std::sync::Arc;
 pub use http_client::{Body, HttpClient, Request, Response};
 
 pub mod logger;
+pub mod redirect;
```
yoshuawuyts, line 58: In Tide we only expose `tide::Redirect` -- I think it would make sense for us to do the same here.

src/middleware/redirect/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+/// # async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+/// let mut res = surf::get("https://httpbin.org/get")
+///     .middleware(surf::middleware::logger::new())
+///     .await?;
+/// dbg!(res.body_string().await?);
+/// # Ok(()) }
+/// ```
+pub fn new() -> Redirect {
+    Redirect::new(3)
+}
```
yoshuawuyts, line 48: If the `redirect` submodule isn't public I don't think we need this method.

src/middleware/redirect/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+
+/// Log each request's duration.
```
yoshuawuyts, line 50: Think you missed this.

src/middleware/redirect/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        next: Next<'a>,
+    ) -> BoxFuture<'a, Result<Response>> {
+        Box::pin(async move {
+            let mut redirect_count: u8 = 0;
+
+            // Note(Jeremiah): This is not ideal.
+            //
+            // HttpClient is currently too limiting for efficient redirects.
+            // We do not want to make unnecessary full requests, but it is
+            // presently required due to how Body streams work.
```
yoshuawuyts, line 78: Thanks for writing this comment!

tests/test.rs
```diff
@@ -105,3 +105,17 @@ SOFTWARE.
 
     Ok(())
 }
+
+// TODO(Jeremiah): Re-enable this one httpbin is not broken, or use a mock server.
```
yoshuawuyts, line 109: fair; good to comment out temporarily.

src/middleware/mod.rs
Fishrock123: See also https://github.com/http-rs/surf/pull/198

Fishrock123:
Should this be feature flagged?

passcod:
Am I right in reading that this keep the same request method regardless of the status code? I don't think that's correct: 301/302 should redirect to GET instead.

PR #202, opened 2020-07-24
Update examples, some docs

Fishrock123:
- avoid showing `logger::new()`
- use `#[async_std::main]`
- do away with turbofish `Ok`

### DIFF
--- a/src/middleware/logger/mod.rs
+++ b/src/middleware/logger/mod.rs
@@ -1,16 +1,6 @@
 //! Logging middleware.
 //!
-//! # Examples
-//!
-//! ```no_run
-//! # #[async_std::main]
-//! # async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
-//! let mut res = surf::get("https://httpbin.org/get")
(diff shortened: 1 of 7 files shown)


src/middleware/logger/mod.rs
```diff
@@ -1,16 +1,6 @@
...
-//! ```no_run
-//! # #[async_std::main]
-//! # async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
-//! let mut res = surf::get("https://httpbin.org/get")
-//!     .middleware(surf::middleware::logger::new())
-//!     .await?;
-//! dbg!(res.body_string().await?);
-//! # Ok(()) }
-//! ```
+//! This middleware is used by default unless the `"middleware-logger"` feature is disabled.
```
goto-bus-stop, line 3: I think it's still useful to show the example in case users disable the feature but want to use the logger for a particular client anyways?

Fishrock123: Fwiw it is shown below as `Logger::new()`

PR #208, opened 2020-07-30
Client: make Clone create a new middleware stack

### LINKED ISSUE #205: Different approach to "per-request" middleware
This is an idea assuming that #194 or a similar approach will land. That PR will likely remove per-request middleware.

A different approach could be to have multiple clients sharing the same client backend, but different middleware stacks. For example,
```rust
let base_client = surf::Client::new()
  .middleware(Logger::new())
  .middleware(AuthenticationMiddleware::new());
let retriable_client = base_client.clone()
  .middleware(Retry::new());

// Now use retriable_client.send() for things that you know can be retried, and base_client for other things
```
In this snippet, `client.clone()` would clone the middleware stack and clone the `Arc<dyn HttpClient>`. The Arc-ed resources would be shared between clients, while non-Arc-ed resources are specific to a client.

If surf gets connection pooling support in the future, I would expect that `retriable_client` and `base_client` share the same connection pool, but `base_client` has 2 middlewares and `retriable_client` has 3. Other configuration knobs like timeouts or proxying could be treated in the same way as middlewares (if they are not implemented as middlewares).

If a particular middleware should really only be used for a single request, that could be achieved by creating a temporary client clone with
```rust
base_client.clone()
  .middleware(SomethingUseful::new())
  .send(req)
```

This is similar to the approach taken by okhttp in Java land, where you can add interceptors to a client (functionally identical to middleware) and use [`client.newBuilder()`](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/#customize-your-client-with-newbuilder) to create new client instances based on existing instances while sharing the backing thread pool.

Fishrock123:
This allows an easier way of having Clients with a shared `HttpClient`, but with potentially different middleware.

Closes #205

Requires https://github.com/http-rs/surf/pull/194

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -26,11 +26,25 @@ use http_client::h1::H1Client;
 /// let (str1, str2) = futures_util::future::try_join(res1, res2).await?;
 /// # Ok(()) }
 /// ```
-#[derive(Clone)]
 pub struct Client {
     http_client: Arc<dyn HttpClient>,
     /// Holds the middleware stack.
     middleware: Arc<Vec<Arc<dyn Middleware>>>,
+}
+
+impl Clone for Client {
+    /// Clones the Client.
+    ///
+    /// This copies the middleware stack from the original, but shares
+    /// the `HttpClient` of the original.
+    /// Note that individual middleware in the middleware stack are
+    /// still shared by reference.
+    fn clone(&self) -> Self {
+        Self {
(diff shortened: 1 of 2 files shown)


Fishrock123:
To clarify here: the middleware when cloned is still the same, reference-based individual middleware, but the stack is new, so that changes to the new client's stack are not reflected in any client it was cloned from.

PR #214, opened 2020-08-13
Replace the native-client feature flags in client.rs and request.rs with its component features

rossdylan:
Previously it was not possible to compile surf with just curl enabled.
This was caused by the usage of the native-client feature which by
default includes both curl and wasm backends. This change alters the
cfg options to gate on either wasm-client or curl-client.

NOTE: This doesn't completely fix compiling with just the curl feature, a separate PR will be opened on http-rs/http-client to finish the job.


Fishrock123:
This seems correct to me.

PR #215, opened 2020-08-15
add an optional base_url to Client

jbr:
Encountered a need for this when trying to build a testing library for tide based on surf, but this seems like a generally useful thing when reusing a client.

I'm not sure if this is the right implementation, but it seemed like the least-invasive change to support this functionality

```rust
let mut client = surf::client();
client.set_base_url(Url::parse("http://example.com/api/v1")?);

let posts: Vec<Post> = client.get("/posts.json").recv_json().await?;
let users: Vec<User> = client.get("/users.json").recv_json().await?;
```

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,7 +1,7 @@
 use std::fmt;
 use std::sync::Arc;
 
-use crate::http::Method;
+use crate::http::{Method, Url};
 use crate::middleware::{Middleware, Next};
 use crate::{HttpClient, Request, RequestBuilder, Response, Result};
 
@@ -27,6 +27,7 @@ use http_client::h1::H1Client;
 /// # Ok(()) }
 /// ```
 pub struct Client {
+    base_url: Option<Url>,
     http_client: Arc<dyn HttpClient>,
     /// Holds the middleware stack.
     middleware: Arc<Vec<Arc<dyn Middleware>>>,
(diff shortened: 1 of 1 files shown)


goto-bus-stop:
Yeah I've wanted this too :) is this something that could/should be a middleware? Not sure where we want to draw that line. This apprach seems good too.

jbr:
@goto-bus-stop I agree it would be nice if it were somehow more deeply integrated into surf, like a middleware, but there's no way to make a Url from a path fragment, and we need to have a Url to build a Request currently. We'd have to change http-types Requests to hold onto a String (or http-types `enum Uri { Url(Url), Path(String) }` or something like that) so that we can transform it into a Url at a later step, and that seemed like a much more sweeping change.

goto-bus-stop:
Makes sense. I think I prefer this approach to having an enum like that in http-types :)

src/client.rs
```diff
@@ -450,7 +448,27 @@ impl Client {
     /// # Ok(()) }
     /// ```
     pub fn patch(&self, uri: impl AsRef<str>) -> RequestBuilder {
-        let uri = uri.as_ref().parse().unwrap();
-        RequestBuilder::new(Method::Patch, uri).with_client(self.clone())
+        RequestBuilder::new(Method::Patch, self.url(uri)).with_client(self.clone())
+    }
+
+    /// sets the base path for this client
```
goto-bus-stop, line 454: for consistent doc looks:
```suggestion
    /// Sets the base URL for this client. All request URLs will be relative to this path.
    ///
    /// # Examples
```

goto-bus-stop (approved):
LGTM

Issue #220, opened 2020-09-07
Headers on a http_types::Request are not sent


jbr:
```rust
    let mut request = http_types::Request::new(
        http_types::Method::Get,
        http_types::Url::parse("configure this").unwrap(),
    );

    request.insert_header("some", "header");

    // does not send the header
    surf::Client::new().send(request).await?;

    // does send the header
    http_client::h1::H1Client::new().send(http_request).await?.into()
```

the server does not see the `some: header` header when sending through surf, but it does when sending through http_client. discovered on surf main branch

PR #222, opened 2020-09-07
Request: fix From<http::Request> for Request

Fishrock123:
Into consumes, so just use the existing struct so we don't have to worry about copying literaly everything.

Fixes https://github.com/http-rs/surf/issues/220


jbr (approved):
Fixes local repro of issue

Issue #217, opened 2020-08-16
Use async-h1 for client by default


Fishrock123:
Seems like a good time to discuss this, is this something that would be desirable for 2.0.0?

@yoshuawuyts @goto-bus-stop @dignifiedquire

Fishrock123:
See the PR for further follow-up: https://github.com/http-rs/surf/pull/223

Fishrock123:
timeouts will be done when https://github.com/http-rs/surf/pull/310 lands

PR #223, opened 2020-09-07
Make async-h1 the default client, don't build with wasm by default.

Fishrock123:
See https://github.com/http-rs/surf/issues/217


Fishrock123:
Discussed this on a call with @yoshuawuyts and @jbr.

It seems like there are still too many things missing from the async-h1 client at the time of writing to make this worthwhile for 2.0.

The list I gathered:
- chunked encoding
- connection pooling
- http/2
- compression
- forwarding

Other things that may be less-than-complete:
- redirects (by default?)
- encodings (?)
- TLS support (?)

Fishrock123:
See also https://github.com/orgs/http-rs/projects/2

PR #230, opened 2020-09-14
src: feature flag each backend independently

Fishrock123:
No longer has a magic wasm default, wasm must be specified independently.

Opens room for per-backend optimization (such as Isahc singleton).


Fishrock123:
@yoshuawuyts This now takes `wasm_bindgen` into account!

yoshuawuyts (approved):
LGTM!

PR #231, opened 2020-09-15
RequestBuilder: use shared client internally if available

Fishrock123:
This should help reduce overhead from creating new Isahc clients.

See https://docs.rs/isahc/0.9.8/isahc/struct.HttpClient.html

From Isahc's Docs:
> The client maintains a connection pool internally and is not cheap to create ...

This may be possible without https://github.com/http-rs/surf/pull/230 but was much easier to do with those config feature changes.


yoshuawuyts (approved):
:100::100::100::100::100:

PR #232, opened 2020-09-18
feature changes for wasm support on 2.x

jbr:
this pr does three things:
* updates http-client to 5.0.0
* removes the native-client feature. since using the wasm client requires explicitly specifying a feature for that, the notion of "native-client" isn't relevant anymore. it existed as a magical "no features needed, works across platforms."
* loads no default features on http-client. without this, surf does not build on wasm. otherwise we build "wasm and also the default features," which currently means h1-client with tls, which won't build on wasm.
* adds a wasm check for ci
* fixes wasm build
* fixes h1 client build

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -125,7 +125,7 @@ impl Client {
     /// ```rust
     /// # #[cfg(feature = "curl-client")] {
     /// # use std::sync::Arc;
-    /// # use http_client::isahc::IsahcClient;
+    /// use http_client::isahc::IsahcClient;
     /// let client = surf::Client::with_http_client(Arc::new(IsahcClient::new()));
     /// # }
     /// ```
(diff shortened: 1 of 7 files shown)


src/client.rs
```diff
@@ -123,9 +123,11 @@ impl Client {
     /// # Examples
     ///
     /// ```rust
+    /// # #[cfg(feature = "curl-client")] {
     /// # use std::sync::Arc;
-    /// # use http_client::h1::H1Client;
-    /// let client = surf::Client::with_http_client(Arc::new(H1Client::new()));
+    /// # use http_client::isahc::IsahcClient;
```
Fishrock123, line 128:
```suggestion
    /// use http_client::isahc::IsahcClient;
```

Fishrock123 (approved):
looks good but the one example really doesn't make sense without the `use` directive included.

PR #233, opened 2020-09-19
Add wasm integration tests

jbr:
This PR adds integration tests that make real http requests to httpbin.org from within headless browsers using surf's wasm-client. This provides both regression testing and also is a starting place for more involved testing to ensure that the wasm client works as expected.

I couldn't figure out how to disable default features specifically for these tests, so instead i added a workspace crate that depends on the parent and exists just for running wasm tests. If there's a better way to do this, please suggest!


src/client.rs
```diff
@@ -131,34 +95,24 @@ impl Client {
...
-            middleware: Arc::new(vec![
-                #[cfg(feature = "middleware-logger")]
-                Arc::new(crate::middleware::Logger::new()),
-            ]),
-        }
+            middleware: Arc::new(vec![]),
+        };
+
+        #[cfg(feature = "middleware-logger")]
+        let client = client.with(crate::middleware::Logger::new());
```
Fishrock123, line 105: Why was this moved out to here rather than just being in the vec?

src/client.rs
```diff
@@ -7,37 +7,20 @@ use crate::{HttpClient, Request, RequestBuilder, Response, Result};
...
-use http_client::wasm::WasmClient;
-
-#[cfg(all(
-    feature = "curl-client",
-    not(any(feature = "h1-client", feature = "wasm-client"))
-))]
+#[cfg(feature = "curl-client")]
+use http_client::isahc::IsahcClient as DefaultClient;
+
+#[cfg(feature = "wasm-client")]
```
Fishrock123, line 13:
```suggestion
#[cfg(any(feature = "wasm-client", feature = "wasm_bindgen")]
```

The other lines should be `all(feature = "x", not(feature = "wasm_bindgen"))`.

wasm-test/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "wasm-test"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+surf = {path = "..", default-features = false, features = ["wasm-client"]}
```
Fishrock123, line 7: Does this work without `features = ["wasm-client"]`? i.e. with implicit `wasm_bindgen`?

Fishrock123 (changes requested):

src/client.rs
jbr: Wasm_bindgen enables wasm-client, though

wasm-test/Cargo.toml
jbr: Wasm-bindgen is useless, I think we can remove it. It only applies to the top level crate, which isn't relevant to a library

src/client.rs
jbr: I found it confusing to read a feature flag in that position, and thought other people might also. I think rust-analyzer also complained about it in that position

src/client.rs
jbr: why would anyone ever enable both wasm and one of the other clients? that doesn't make sense, so we don't need to test for that condition throughout the code

Fishrock123: That is unfortunately how cargo features work, though.

```toml
surf = { version = "2", features = ["wasm-client"] }
```
Enables both the defaults (including `curl-client`) and `wasm-client`.

This was an oversight in my last `cfg` adjusting patch, adn https://github.com/http-rs/surf/pull/234 fixes that properly.

jbr: yeah, people need to say default-features = false, and we need to document that. or we could not choose a default. wasm cannot build if other clients are enabled

Fishrock123: I do not think I agree. The `cfg` statements _can_ handle this fallback properly. I went that way originally but I can't see any downside to having the default client deal with feature overrides of the default.

jbr: if the other features are enabled, the dependencies get pulled in. this is not what users want, but more importantly, those deps cannot build on wasm. people need to disable default features in order to use wasm (and i think we should encourage people to disable default features to switch to h1, as well, in order to not bloat their deps with unused crates)

src/client.rs
jbr: what's the argument against using `with`? another argument for it is that if the logic for what needed to be done in order to insert a middleware changed, there would only be one place that had to change

jbr: This is also how tide does default middleware: https://github.com/http-rs/tide/blob/main/src/server.rs#L100-L102

src/client.rs
Fishrock123: Hmm, true. Ok.

Fishrock123 (approved):

PR #235, opened 2020-09-19
remove wasm_bindgen feature (retain wasm-client, which is identical in practice)

jbr:
As mentioned in the review of #233, I think the wasm_bindgen feature was not the right path for surf, as a library crate. It is only added by wasm_pack to the top level crate, which isn't relevant to a library crate, so users will still need to explicitly specify a wasm feature. As long as we require specifying a wasm feature, we should: a) only have one wasm flag, instead of two identical flags, and b) use a flag that is descriptive (wasm-client matches with h1-client and curl-client)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,6 @@ default = ["curl-client", "middleware-logger", "encoding"]
 h1-client = ["http-client/h1_client"]
 curl-client = ["http-client/curl_client", "once_cell"]
 wasm-client = ["http-client/wasm_client"]
-wasm_bindgen = ["wasm-client"]
 middleware-logger = []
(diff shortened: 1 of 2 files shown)


Fishrock123 (approved):

Fishrock123:
(Maybe this is a bug in wasm-pack though - shouldn't it pass it to all crates? (if that is possible?))

jbr:
I don't think there's a way to pass features into nested / transitive deps. That would be cool, but it doesn't seem to be happening (or we wouldn't need to enable any features for the tests to pass in the integration test PR)

Fishrock123:
I honestly don't know, I wasn't sure if wasm-pack did some kind of special underhand magic.

PR #236, opened 2020-09-21
Allow surf to build without a default client

jbr:
This PR:
* ensures that surf can build as much as possible without an enabled client.  This is important in use cases where someone is providing their own HttpClient backend, as supported by surf features.
* simplifies the logic for checking if there's a client by adding a "default-client" feature that's enabled by all of the other clients.  This will simplify addition of future client backends.
* introduces the use of cfg_if, in order to make the mutually-exclusive conditional logic more clear.

Not covered by this PR: Considering what happens if someone enables _multiple_ backends.

There are exactly two places that panic in this PR that would not have panicked otherwise.  We could return a custom error if that's preferred, although it seems like "trying to send a request without a default client backend and without an explicitly configured http_client" is exactly the sort of developer-caused error scenario that panics are ideal for. This also should be a fairly rare condition, because anyone using surf without default features likely understands that they can't rely on a default client.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,9 +20,10 @@ edition = "2018"
 # when the default feature set is updated, verify that the `--features` flags in
 # `.github/workflows/ci.yaml` are updated accordingly
 default = ["curl-client", "middleware-logger", "encoding"]
-h1-client = ["http-client/h1_client"]
-curl-client = ["http-client/curl_client", "once_cell"]
-wasm-client = ["http-client/wasm_client"]
+h1-client = ["http-client/h1_client", "default-client"]
+curl-client = ["http-client/curl_client", "once_cell", "default-client"]
+wasm-client = ["http-client/wasm_client", "default-client"]
+default-client = []
 middleware-logger = []
 # requires web-sys for TextDecoder on wasm
 encoding = ["encoding_rs", "web-sys"]
@@ -41,6 +42,7 @@ async-std = { version = "1.6.0", default-features = false, features = ["std"] }
 async-trait = "0.1.36"
 pin-project-lite = "0.1.1"
 once_cell = { version = "1.4.1", optional = true }
+cfg-if = "0.1.10"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 # encoding
--- a/src/request_builder.rs
+++ b/src/request_builder.rs
@@ -224,13 +224,11 @@ impl RequestBuilder {
     }
 
     /// Create a `Client` and send the constructed `Request` from it.
-    pub fn send(self) -> BoxFuture<'static, Result<Response>> {
-        if let Some(client) = self.client {
-            client.send(self.req.unwrap()) // Ownership nonsense, same as `build()`.
-        } else {
-            let client = Client::new_shared();
-            client.send(self.build())
-        }
+    pub fn send(mut self) -> BoxFuture<'static, Result<Response>> {
+        self.client
+            .take()
+            .unwrap_or_else(Client::new_shared_or_panic)
+            .send(self.build())
     }
 }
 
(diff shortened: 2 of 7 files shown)


src/request_builder.rs
```diff
@@ -224,13 +224,11 @@ impl RequestBuilder {
...
-            client.send(self.req.unwrap()) // Ownership nonsense, same as `build()`.
-        } else {
-            let client = Client::new_shared();
-            client.send(self.build())
-        }
+    pub fn send(mut self) -> BoxFuture<'static, Result<Response>> {
+        self.client
+            .take()
+            .unwrap_or_else(Client::new_shared_or_panic)
+            .send(self.build())
```
Fishrock123, line 231: Oh, this is nicer!

src/request_builder.rs
```diff
@@ -246,11 +244,10 @@ impl Future for RequestBuilder {
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         if self.fut.is_none() {
             let req = self.req.take().unwrap();
-
             if let Some(client) = &self.client {
                 self.fut = Some(client.send(req))
             } else {
-                let client = Client::new_shared();
+                let client = Client::new_shared_or_panic();
```
Fishrock123, line 250: It might be too much cfg statements but `RequestBuilder` would _ideally_ not impl future if there isn't a default.

Cargo.toml
```diff
@@ -20,9 +20,10 @@ edition = "2018"
 # when the default feature set is updated, verify that the `--features` flags in
 # `.github/workflows/ci.yaml` are updated accordingly
 default = ["curl-client", "middleware-logger", "encoding"]
-h1-client = ["http-client/h1_client"]
-curl-client = ["http-client/curl_client", "once_cell"]
-wasm-client = ["http-client/wasm_client"]
+h1-client = ["http-client/h1_client", "default-client"]
+curl-client = ["http-client/curl_client", "once_cell", "default-client"]
+wasm-client = ["http-client/wasm_client", "default-client"]
+default-client = []
```
Fishrock123, line 26: Since this is not an actual feature I suggest doing some `#[cfg_attr()]` in `lib.rs` to set `#[cfg(default-client)]` instead, maybe?

Fishrock123 (approved):
Overall this seems good but could perhaps be improved a bit still.

src/request_builder.rs
Fishrock123: If this code didn't exist this could probably point to `Client::default()` which could use a shared client if available, changing this to `.unwrap_or_default()`, and eliminating `new_shared()` in the process? Maybe that would be confusing as it would differ from `new()` though.

src/request_builder.rs
jbr: That would interfere with this usage:

```rust
/// surf with no default features
Client::with_http_client(Arc::new(some_http_client)) // Client
    .get("http://example.com") // RequestBuilder
    .await // using Future on RequestBuilder
```

Cargo.toml
jbr: I tried to do this, but nothing I find in the docs indicates it works and I can't find an invocation of cfg_attr that actually conditionally sets a new configuration flag like that. I checked out the hyper branch in order to try to learn from it, but I'm also not convinced it's doing what we want in that branch — if you add a random line like `const S: &str = "";` after the cfg_attr line but before the next cfg attribute macro, it doesn't compile.

I've seen other crates using undocumented cargo.toml features like this, so my guess is that this is still the way to do conditional features

Fishrock123: Huh, upon examining https://doc.rust-lang.org/reference/conditional-compilation.html it seems I am misusing it in that PR. I have no idea how that PR manages to compile...

PR #238, opened 2020-09-24
client: make with_http_client() generic

Fishrock123:
This avoids leaking out the storage underneath.

Internally we require a function like the original though, due to very confusing & indirect lifetime requirements.

I didn't like the api here to begin with, so this seems very reasonable to get in to 2.0


jbr (approved):
Looks like an improvement to me!

PR #240, opened 2020-09-24
deps, client: update for HttpClient 6.0

Fishrock123:
All of the global shared client now uses just one Arc.

Refs: https://github.com/http-rs/http-client/pull/48


Fishrock123:
2.0 milestone due to `httpClient` semver upgrade

PR #243, opened 2020-09-26
remove url crate re-export

yoshuawuyts:
The url crate is part of http-types; there is no need to re-export the full submodule from here. This cleans up the module index slightly. Thanks!


goto-bus-stop:
looks like it needs a couple of docs updates

Fishrock123:
superseded by https://github.com/http-rs/surf/pull/249

PR #246, opened 2020-10-01
RequestBuilder: add .query()

Fishrock123:
This was missing form the original RequestBuilder. My bad.

This would be very handy for me, so I'm pretty eager to get it in asap.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,6 @@ log = { version = "0.4.7", features = ["kv_unstable"] }
 mime_guess = "2.0.3"
 serde = "1.0.97"
 serde_json = "1.0.40"
-serde_urlencoded = "0.6.1"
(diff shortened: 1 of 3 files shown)


jbr (approved):

jbr:
Side note: In reading through this, I ran into this (somewhat-unrelated) line: https://github.com/http-rs/surf/blob/main/src/request.rs#L129 — any idea what's going on there? we've already mutated the url, why are we cloning it?

Fishrock123:
Huh. Good question. b6113dec3df05995dc4d5506c4dd81db4a3e7d9b is the answer but I can't think why I left that stuff.

I've removed `serde_urlencoded` entirely in favor of just using http-types' querystring impl (which uses `serde_qs`).

yoshuawuyts (approved):

PR #250, opened 2020-10-06
Client: propagate base_url through middleware stack

Fishrock123:
This was missed from ad2be972da

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -193,9 +193,15 @@ impl Client {
             })
         });
 
-        let res = next
-            .run(req, Self::with_http_client_internal(http_client))
-            .await?;
+        let client = Self {
+            base_url: self.base_url.clone(),
+            http_client,
+            // Erase the middleware stack for the Client accessible from within middleware.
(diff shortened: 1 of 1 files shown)


src/client.rs
```diff
@@ -193,9 +193,15 @@ impl Client {
...
-        let res = next
-            .run(req, Self::with_http_client_internal(http_client))
-            .await?;
+        let client = Self {
+            base_url: self.base_url.clone(),
+            http_client,
+            // Erase the middleware stack for the Client accessible from within middleware.
+            // This avoids gratuitous circular borrow & logic issues.
+            middleware: Arc::new(vec![]),
+        };
```
Fishrock123, line 202: I think just creating like this is probably the cleanest/clearest about what is happening, and helps prevent us from missing any future addition.

yoshuawuyts (approved):
Looks good. Would a regression test be useful here?

Fishrock123:
On second look, I don't think a regression test is necessary. Fully re-constructing the struct there should already catch anything missing. It forces us to think about this consciously on implementation.

PR #252, opened 2020-10-13
Fix client base URL documentation

CrockAgile:
The client base URL example sets the base URL to
"http://example.com/api/v1", which is then joined with a URI of
"/posts.json". This had two bugs:

- The base url did not end with '/', and so 'v1' was treated as a file
path, and dropped from the result. This lead to resulting URLs like:
"http://example.com/api/posts.json", without the 'v1'.
- the joined URI (/posts.json) begins with a '/', and so it was treated as an absolute
path from the domain. This lead to resulting URLs like:
"http://example.com/posts.json", without the '/api/v1'.

This commit fixes both of these bugs in the example. It also adds a note
to the documentation (copied from the URL crate) about the importance of
the ending '/'.

One test was added to ensure the base_url behavior.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -501,13 +501,17 @@ impl Client {
 
     /// Sets the base URL for this client. All request URLs will be relative to this URL.
     ///
+    /// Note: a trailing slash is significant.
+    /// Without it, the last path component is considered to be a “file” name
+    /// to be removed to get at the “directory” that is used as the base.
+    ///
     /// # Examples
     /// ```no_run
     /// # use http_types::Url;
     /// # fn main() -> http_types::Result<()> { async_std::task::block_on(async {
     /// let mut client = surf::client();
-    /// client.set_base_url(Url::parse("http://example.com/api/v1")?);
-    /// client.get("/posts.json").recv_json().await?; /// http://example.com/api/v1/posts.json
+    /// client.set_base_url(Url::parse("http://example.com/api/v1/")?);
+    /// client.get("posts.json").recv_json().await?; /// http://example.com/api/v1/posts.json
     /// # Ok(()) }) }
     /// ```
     pub fn set_base_url(&mut self, base: Url) {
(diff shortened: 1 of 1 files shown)


src/client.rs
```diff
@@ -522,3 +526,17 @@ impl Client {
...
+    use super::Client;
+    use crate::Url;
+
+    #[test]
+    fn base_url() {
+        let mut client = Client::new();
+        client.set_base_url(Url::parse("http://example.com/api/v1/").unwrap());
+        let url = client.url("posts.json");
+        assert_eq!(url.as_str(), "http://example.com/api/v1/posts.json");
+    }
```
CrockAgile, line 541: I debated putting this assertion in the documentation test, but they were all already marked as `no_run`. Just let me know where you prefer!

jbr:
Thanks for catching this! My inclination would be to treat this as a bug in the code instead of poor documentation. Is there any reason the trailing slash *should* be important? We should add one if the base url doesn’t end with /

CrockAgile:
Yeah, this being a bug is a possible interpretation! I'm by no means an expert on the WHATWG URL standard tho, and this behavior arises from the canonical [servo/rust-url](https://github.com/servo/rust-url) crate. Reading over some of their issues, it seems this kind of behavior is a little confusing/ambiguous.

The URL crate was made for use in servo, and it seems both Chrome and Firefox follow this behavior:

```js
// base URL missing trailing slash
(new URL('posts.json', 'http://example.com/api/v1')).toString()
"http://example.com/api/posts.json"

// relative path beginning with slash
(new URL('/posts.json', 'http://example.com/api/v1/')).toString()
"http://example.com/posts.json"

// "intended" result
(new URL('posts.json', 'http://example.com/api/v1/')).toString()
"http://example.com/api/v1/posts.json"
```

So if that's how joining URL and a path "should" work, what options does surf have?
- enforce that the "base URL" ends with a "/" by:
  - appending a "/" if none
  - throwing an exception if none
- keeping as is, but highlighting URL weirdness in the docs
- ???

CrockAgile:
The "???" was meant to leave room for any other proposals anyone has, so please speak up if you see a good way to handle this.

Fishrock123 (approved):

Fishrock123:
I've created an issue for this at https://github.com/http-rs/surf/issues/255

I'm going to merge this as a stop-gap for now.

PR #263, opened 2020-11-04
ci improvements

Fishrock123:
In particular, also run ci on the `"hyper-client"` backend, and don't build docs for deps.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
         rust: [nightly]
-        backend: [curl-client, h1-client]
+        backend: [curl-client, h1-client, hyper-client]
 
     steps:
     - uses: actions/checkout@master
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -71,17 +71,16 @@ jobs:
...
     - uses: actions/checkout@master
-    - uses: actions-rs/toolchain@v1
-      with:
-        toolchain: nightly
-        components: rustfmt, clippy
-        override: true
+
+    - name: nightly setup
+      run: rustup toolchain install nightly
+      run: rustup default nightly
```
goto-bus-stop, line 77: duplicate `run` key doesn't look right?

goto-bus-stop:
LGTM once that ^ fixed 👍

.github/workflows/ci.yaml
Fishrock123: No this is right, we could also do `cargo +nightly <command>` for the commands below.

goto-bus-stop: sorry I sholuld've been clearer, the command looks right, but the duplicate `run` keys are not valid YAML

Fishrock123: Oh! Huh, I wonder why this worked at all then.

goto-bus-stop: i dont think it did 😝 https://github.com/http-rs/surf/actions/runs/346519413

Fishrock123: It's because it ran no jobs, i.e. failed entirely.

Fishrock123:
Updated, improved even a bit more. Finally merging this.

I have reverted the change to no longer use nightly for `fmt, clippy, docs` as it turns out that actions-rs/toolchain does indeed do downgrades if components are not found, or at least, it says it does: https://github.com/actions-rs/toolchain#components

PR #265, opened 2020-11-17
Add native-client feature

MarcAntoine-Arnaud:


Fishrock123:
Why?

MarcAntoine-Arnaud:
> Why?

To select backend with the `native-client` for every target, and don't need to play with features for targets...

Issue #261, opened 2020-11-03
Per-request middleware? (in RequestBuilder)


aldanor:
(somewhat related to #259)

Would be nice to be able to do something like `request_builder.with(my_middleware)` which would replace the client it holds with a version that has the given middleware pushed on the stack (using the default shared client as the base if needed).

This would also make it possible to apply middleware to the default shared client on a per-request basis (which is arguably not the most efficient way to do so, but still):

```rust
surf::get("http://...").with(...)
```

Fishrock123:
I think this is unlikely to happen - see https://github.com/Fishrock123/surf/pull/1

Since clients share underlying resources anyways, you should just clone your client, or use one if you aren't.

Fishrock123:
> This would also make it possible to apply middleware to the default shared client on a per-request basis (which is arguably not the most efficient way to do so, but still):
> 
> ```rust
> surf::get("http://...").with(...)
> ```

If you are doing advanced usage already, you should probably not do this. It will always be less efficient somehow (even if that is just creating new memory for middleware to sit in). IMO, we should recommend using clients, and improve that story as much as possible.

Fishrock123:
Stated differently: It would be nice to see an argument for cases where this is borderline necessary.

aldanor:
@Fishrock123 Thanks for taking time to answer!

I agree that promoting client usage as much as possible is a good idea (although it's currently not clear from the docs, and the first block of code you see in the readme uses `surf::get` as well). Although there will always be cases where you don't really care much about performance and prefer simpler API, I guess (like in the examples in the readme).

One example of a use case you've asked for would be more or less like this: you have a single client that holds connection pool, and your application makes multiple types/group of requests - e.g. A, B and C - where, say, you want to log requests A, and you also want to validate and log requests of type C in a different way, and requests B you don't care about. This would currently require you to either:

- Write one middleware that's supposed to know "everything about everyone"; i.e. in the example above, A, B and C. This doesn't scale too well as you keep extending the logic and the number of request types and groups.
- Clone the client every time before making a request? Just for the sake being use a different middleware with it? It's not obvious how cheap or expensive it is cloning a client and what does it do behind the curtain for each of the client types; at least it's not clearly documented. Either way, it seems like a bit of overkill, especially if you make many requests.
[...]

PR #266, opened 2020-11-21
feat: add per-request middleware

Fishrock123:
I think Tide & Surf's focus on ergonomics predicates this should happen, even if it's not the most efficient.
Were it for efficiency alone we would not have the top-level http-verb quick methods.
Hopefully the notice in the docs will put people in the correct direction for most cases.

This adds an optional middleware stack to `Request`, allowing `RequestBuilder` and one-off requests to have middleware added onto them, but also allows one-off requests with middleware to have the middleware stack be appended to client middleware for just that request.

Closes https://github.com/http-rs/surf/issues/261

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -177,6 +177,8 @@ impl Client {
 
     /// Send a `Request` using this client.
     ///
+    /// Client middleware is run before per-request middleware.
+    ///
     /// # Examples
     ///
     /// ```no_run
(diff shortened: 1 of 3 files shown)


src/client.rs
```diff
@@ -177,6 +177,8 @@ impl Client {
 
     /// Send a `Request` using this client.
     ///
+    /// Client middleware is run before per-request middleware.
```
yoshuawuyts, line 180: nice; I think this clarification will be helpful

yoshuawuyts (approved):
This looks really good! I think all this PR is missing is a test to ensure the per-request middleware is constructed correctly, and does not accidentally modify the client middleware. But :100: work!

PR #271, opened 2020-12-07
Add 'h1-client-rustls' feature relying on rustls

JEnoch:
This PR is a follow-up of http-rs/http-client#53 and addresses #40.
It adds a `h1-client-rustls` feature using `http-client/h1_client_rustls` feature introduced by http-rs/http-client#53.

**TODO**: 
For testing purpose I set the dependency to `http-client` to directly use its Git repo.
Once http-client is released with http-rs/http-client#53, update the dependency to use the latest version.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
         rust: [nightly]
-        backend: [curl-client, h1-client, hyper-client]
+        backend: [curl-client, 'h1-client native-tls', hyper-client]
 
     steps:
     - uses: actions/checkout@master
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
         rust: [nightly]
-        backend: [curl-client, h1-client, hyper-client, h1-client-rustls]
+        backend: [curl-client, h1-client, hyper-client]
 
     steps:
     - uses: actions/checkout@master
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
         rust: [nightly]
-        backend: [curl-client, h1-client, hyper-client]
+        backend: [curl-client, h1-client, hyper-client, h1-client-rustls]
 
     steps:
     - uses: actions/checkout@master
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,9 +22,10 @@ edition = "2018"
 default = ["curl-client", "middleware-logger", "encoding"]
 curl-client = ["http-client/curl_client", "once_cell", "default-client"]
 h1-client = ["http-client/h1_client", "default-client"]
-h1-client-rustls = ["http-client/h1_client_rustls", "default-client"]
 hyper-client = ["http-client/hyper_client", "once_cell", "default-client", "async-std/tokio02"]
 wasm-client = ["http-client/wasm_client", "default-client"]
+native-tls = ["http-client/native-tls"]
+rustls = ["http-client/rustls"]
 default-client = []
 middleware-logger = []
 # requires web-sys for TextDecoder on wasm
@@ -36,8 +37,7 @@ log = { version = "0.4.7", features = ["kv_unstable"] }
 mime_guess = "2.0.3"
 serde = "1.0.97"
 serde_json = "1.0.40"
-# http-client = { version = "6.1.0", default-features = false }
-http-client = { git = "https://github.com/http-rs/http-client.git", branch = "main", default-features = false }
+http-client = { version = "6.3.0", default-features = false }
 http-types = "2.5.0"
 async-std = { version = "1.6.0", default-features = false, features = ["std"] }
 async-trait = "0.1.36"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ edition = "2018"
 default = ["curl-client", "middleware-logger", "encoding"]
 curl-client = ["http-client/curl_client", "once_cell", "default-client"]
 h1-client = ["http-client/h1_client", "default-client"]
+h1-client-rustls = ["http-client/h1_client_rustls", "default-client"]
 hyper-client = ["http-client/hyper_client", "once_cell", "default-client", "async-std/tokio02"]
 wasm-client = ["http-client/wasm_client", "default-client"]
 default-client = []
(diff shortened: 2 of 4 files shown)


JEnoch:
CI failures for `curl-client` feature are caused by http-rs/http-client#57.
The workflow should be re-started after merge of http-rs/http-client#58

Fishrock123:
I will look more at the failures soon, but I think this would perhaps be better named as `"h1-rustls-client"`?

Ideally I'd like to separate out the tls backend flags... hyper has the same thing, I think. I haven't been able to figure out a good way to do that though...

JEnoch:
@Fishrock123 I don't see such separation in [hyper's features](https://github.com/hyperium/hyper/blob/master/Cargo.toml#L71).
And I'm not sure how to achieve this:
 - if `h1-client` feature is selected, `http-client/h1_client` feature is used, that depends on `async-native-tls` (and thus on OpenSSL)
 - but if in addition a `rustls` feature is selected:
     - either the `http-client/h1_client` has to be replaced with `http-client/h1_client_rustls`
     - either you keep the `http-client/h1_client` usage, adding a usage of a new `http-client/rustls` feature that would replace the dependency to `async-native-tls` with a dependency to `async-tls`

I don't think Cargo provides a way to declare such _"feature replacement in case of another feature usage"_.

Thinking further, a solution could be such reorg in http-client features:
```toml
[features]
default = ["h1_client", "async-native-tls"]
h1_client = ["async-h1", "async-std"]
h1_rustls = ["async-tls"]
```
And re-exposing the same in Surf.

A user wanting h1 + rustls would use such dependency:
```toml
surf = { version = "2.1.0", default-features = false, features = ["h1-client", "h1-rustls"] }
```
The drawback is that using `h1-client` as only feature will fail to build since a TLS impl is missing. This would be a regression wrt. current behaviour.

What do you think ?

Fishrock123:
> I don't see such separation in hyper's features.

[`reqwest`](https://crates.io/crates/reqwest), the hyper client wrapper, has this through. Their feature flags are pretty comprehensive. Maybe there's something to learn from them:
https://github.com/seanmonstar/reqwest/blob/bd9ff9f371b756decf26fbbde1687433b0f63774/Cargo.toml#L30-L41

Fishrock123:
> Thinking further, a solution could be such reorg in http-client features:

I also think we should re-organize http-client features. It's a big reason i never did another http-client release yet, because I wasn't sure what to do.

> The drawback is that using `h1-client` as only feature will fail to build since a TLS impl is missing. This would be a regression wrt. current behaviour.

This actually isn't so bad, we can make http-client compile without tls support if nothing is provided.

Ideally, I think, http-client would have `native-tls` and `rustls` features, but this is a problem because cargo has no "join" on features e.g. make `h1_client`+`native-tls` pull in something different than maybe `curl_client`+`native-tls`. But maybe this isn't an issue if both hyper and isahc just use the same tls implementations as us anyways?

Fishrock123:
I have created a zulip thread about this to discuss with the cargo team... https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/co-dependant.20feature.20flag.20dependencies

Fishrock123:
Maybe we just need to stick with `h1-client-rustls` for now though... this is also blocking https://github.com/nlopes/libhoney-rust/pull/61

Fishrock123:
@JEnoch mind to rebase this?

Fishrock123:
It seems like https://github.com/rust-lang/cargo/issues/8832 will eventually allow `h1-client`+`rustls-tls` to work, but no timeline on when.

JEnoch:
> @JEnoch mind to rebase this?

Sure! Doing this shortly.

JEnoch:
Note that this PR [relies on main branch of http-client](https://github.com/http-rs/surf/pull/271/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R40) (for http-rs/http-client#53 and http-rs/http-client#58).
You might want to release http-client before merging this PR. Just let me know and I'll update the Cargo.toml.

.github/workflows/ci.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
         rust: [nightly]
-        backend: [curl-client, h1-client, hyper-client]
+        backend: [curl-client, h1-client, hyper-client, h1-client-rustls]
```
Fishrock123, line 22: I'd rather test this in http-client, to avoid ballooning the feature matrix too much here

Cargo.toml
```diff
@@ -22,6 +22,7 @@ edition = "2018"
 default = ["curl-client", "middleware-logger", "encoding"]
 curl-client = ["http-client/curl_client", "once_cell", "default-client"]
 h1-client = ["http-client/h1_client", "default-client"]
+h1-client-rustls = ["http-client/h1_client_rustls", "default-client"]
```
Fishrock123, line 25: I have a PR up on http-client, I'd like to separate the feature flag out there first even if we can't here: https://github.com/http-rs/http-client/pull/67

Fishrock123 (changes requested):

.github/workflows/ci.yaml
JEnoch: Sure, I'll remove it.

Cargo.toml
JEnoch: So, once you merge it in http-client, would this PR still make sense ?
I can either adapt this PR to have:
```toml
h1-client = ["http-client/h1_client", "http-client/native-tls", "default-client"]
h1-client-rustls = ["http-client/h1_client", "http-client/rustls", "default-client"]
```

Or you can either re-expose the same `native-tls` and `rustls` in surf:
```toml
h1-client = ["http-client/h1_client", "default-client"]
native-tls = ["http-client/native-tls"]
rustls = ["http-client/rustls"]
```
Meaning that user wanting h1-client would have to use the combination of `h1-client` + `native-tls` or `rustls `.

What do you prefer ?

Fishrock123: You know what, since we only support this on h1 client anyways (due the lack of `rustls` support in `hyper-tls`), let's separate these and then whenever https://github.com/rust-lang/cargo/issues/8832 exists and `hyper-tls` supports `rustls` we can worry about that then, since this is the ideal path.

Fishrock123:
I've released http-client 6.3.0

Cargo.toml
JEnoch: Done:
 - Bump http-client to 6.3.0
 - Removed `h1-client-rustls` feature and added `native-tls` + `rustls`
 - Adapted ci.yaml since using `h1-client` feature only doesn't work anymore
 - Updated inline doc for features in src/lib.rs

Let me know if it's fine for you!

Fishrock123:
Merged a working set of features as https://github.com/http-rs/surf/pull/292, which is more like the original without my poor suggestion on feature separation (which is not yet supported by cargo in the way which we'd need).

Issue #276, opened 2021-01-12
Trailing slash redirects


hipstermojo:
When using the default redirect middleware on surf, I noticed that it exits early when the next redirected URL is invalid. However, I think this is recoverable if the next location is a relative URL. This was the case when surf requests a URL without a trailing slash and the server redirects to a URL with a trailing slash. I made a fix for it and wanted to know if it is worth making a PR for. I saw that something similar was mentioned in #252.

Fishrock123:
@hipstermojo Please send a PR

PR #279, opened 2021-01-31
Add fix for redirect middleware for #276

hipstermojo:
This fix will require a clone of the original base URL which can then be joined to the value in the `LOCATION` header if it is a relative path. Previously this would simply return an Error for RelativeURLWithoutBase.

An example of such a URL is https://blog.thomasheartman.com/posts/on-generics-and-associated-types which would redirect to add a trailing slash.

### DIFF
--- a/src/middleware/redirect/mod.rs
+++ b/src/middleware/redirect/mod.rs
@@ -91,6 +91,8 @@ impl Middleware for Redirect {
         // and try sending it until we get some status back that is not a
         // redirect.
 
+        let base_url = req.url().clone();
+
         while redirect_count < self.attempts {
             redirect_count += 1;
             let r: Request = req.clone();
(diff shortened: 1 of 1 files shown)


src/middleware/redirect/mod.rs
```diff
@@ -91,14 +91,24 @@ impl Middleware for Redirect {
...
         while redirect_count < self.attempts {
             redirect_count += 1;
             let r: Request = req.clone();
             let res: Response = client.send(r).await?;
             if REDIRECT_CODES.contains(&res.status()) {
                 if let Some(location) = res.header(headers::LOCATION) {
                     let http_req: &mut http::Request = req.as_mut();
-                    *http_req.url_mut() = Url::parse(location.last().as_str())?;
+                    *http_req.url_mut() = match Url::parse(location.last().as_str()) {
+                        Ok(valid_url) => valid_url,
```
Fishrock123, line 104: This is not quite complete, a redirect could give us a full url and then a future one could give us a partial one.

I think this branch should also update the`base_url`.

Fishrock123 (changes requested):

hipstermojo:
The linting issues with Clippy are technically outside the scope of this PR so I'm not sure if I should edit that part or not.

Fishrock123 (approved):
Could you rebase so that CI updates? thanks, looks good otherwise.

Fishrock123:
Merged the exact same commits in https://github.com/http-rs/surf/pull/293 - for some reason there was still a rebase issue here.

Thanks!!

PR #292, opened 2021-03-01
Add 'h1-client-rustls' feature relying on rustls

Fishrock123:
This PR is a follow-up of http-rs/http-client#53 and addresses #40.
It adds a `h1-client-rustls` feature using `http-client/rustls` feature introduced by http-rs/http-client#53.

Co-Authored-By: Julien Enoch <julien.enoch@adlinktech.com>

----

Successor to https://github.com/http-rs/surf/pull/271 so I can hopefully merge and release today. That PR took changes on which were my own suggestion but which do not actually work correctly (no default tls support) for which I am sorry...

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
         rust: [nightly]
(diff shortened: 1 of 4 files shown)


Fishrock123:
... is this ALSO https://github.com/http-rs/http-client/issues/71?

PR #302, opened 2021-04-29
Add specialized body types to `RequestBuilder`

yoshuawuyts:
I was doing a coding exercise and realized that both `surf::RequestBuilder` and `tide::ResponseBuilder` didn't have body shorthand methods yet. This makes it so `surf::RequestBuilder` is closer to `surf::Request`, making it easier to write. Thanks!

## Example

```rust
let uri = "https://httpbin.org/post";
let data = &Ip { ip: "129.0.0.1".into() };

// before
let res = surf::post(uri).body(Body::from_json(data)?).await?;

// after
let res = surf::post(uri).body_json(data)?.await?;
```

### DIFF
--- a/src/request_builder.rs
+++ b/src/request_builder.rs
@@ -120,7 +120,10 @@ impl RequestBuilder {
         self
     }
 
-    /// Sets the body of the request.
+    /// Pass an `AsyncRead` stream as the request body.
+    /// # Mime
+    ///
+    /// The encoding is set to `application/octet-stream`.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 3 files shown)


src/request_builder.rs
```diff
@@ -120,7 +120,10 @@ impl RequestBuilder {
         self
     }
 
-    /// Sets the body of the request.
+    /// Pass an `AsyncRead` stream as the request body.
```
Fishrock123, line 123: Maybe anything with implements `<link to body docs Into>`, ... "such as `AsyncRead`"?

src/request_builder.rs
```diff
@@ -120,7 +120,10 @@ impl RequestBuilder {
...
+    /// # Mime
+    ///
+    /// The encoding is set to `application/octet-stream`.
```
Fishrock123, line 126:
```suggestion
    /// The `content-type` is set to `application/octet-stream`.
```

& similar spots

(To not confuse with other "encoding" headers.)

src/request_builder.rs
```diff
@@ -138,6 +141,108 @@ impl RequestBuilder {
...
+    /// ```no_run
+    /// # #[async_std::main]
+    /// # async fn main() -> surf::Result<()> {
+    /// let uri = "https://httpbin.org/post";
+    /// let res = surf::post(uri).body_file("./archive.tgz").await?.await?;
+    /// assert_eq!(res.status(), 200);
+    /// # Ok(()) }
+    /// ```
+    #[cfg(not(target_arch = "wasm32"))]
+    pub async fn body_file(self, path: impl AsRef<std::path::Path>) -> std::io::Result<Self> {
```
Fishrock123, line 242: This should probably be something more like `body_path`, so that we can take `body_file` from a `File` handle - see https://github.com/http-rs/http-types/pull/318

Fishrock123 (changes requested):

src/request_builder.rs
yoshuawuyts: I agree with that, but that would also make it inconsistent with `Request::body_file`. Perhaps we should add this as-is now, and flag it as a change for the next major?

src/request_builder.rs
yoshuawuyts: Not sure I quite follow -- can you maybe share what text you'd like to see here instead?

src/request_builder.rs
Fishrock123: Yeah we could do that

src/request_builder.rs
Fishrock123:
```suggestion
    /// Sets the body of the request from any type with implements `Into<Body>`, for example, any type with is `AsyncRead`.
```
Perhaps?

Fishrock123 (approved):
Feel free to address the first comment if you think that text is suitable; else this seems good. I've certainly run into this myself!

Issue #260, opened 2020-11-03
Client -> RequestBuilder with a user-provided Method?


aldanor:
It looks like this has been oversought. How do you create a `RequestBuilder` from a `Client` (which may have a base url and the corresponding url logic) if you have a `Method` and an url?

Maybe something like `client.request(Method::Get, "foo")` would be worthwhile adding?

(There's `Client::send()`, but perhaps there might be a more ergonomic option)

jbr:
I've run into this as well and think that we can do better than `client.send(Request::new(Method::Get, Url::parse("http://_").unwrap()).await?`

aldanor:
@jbr Moreover, I think that's not the same thing because of the whole `base_url` logic? (this part is completely unobvious, as a matter of fact)

jbr:
That's true — it diverges if you're using base_url

Fishrock123:
I agree, this appears to be an oversight. 

I'd take a pr for `Client::request(&self, Method, Url) -> RequestBuilder`.

Fishrock123:
Note: once the change described in is made, this would probably change to:
```
Client::request(&self, Method) -> RequestBuilder
```

Since `Client` would normally have a base_url set already. If this form was still needed a `request_with_url` could be introduced, although I doubt that would be necessary.

utsavm9:
Hey! Let me know if #308 is the intended function we want to add.

Fishrock123:
Done in #308

PR #308, opened 2021-06-25
RequestBuilder from Client and user-provided method

utsavm9:
Let me know if this change was intended in the conversation from https://github.com/http-rs/surf/issues/260. 

Tested by just running the code sample in the doc. There do not seem to be unit tests for other similar client methods and so this also does not have any.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -523,6 +523,29 @@ impl Client {
         RequestBuilder::new(Method::Patch, self.url(uri)).with_client(self.clone())
     }
 
+    /// Perform a HTTP request with the given verb using the `Client` connection.
+    ///
+    /// # Panics
+    ///
+    /// This will panic if a malformed URL is passed.
+    ///
+    /// # Errors
+    ///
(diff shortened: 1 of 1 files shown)


src/client.rs
```diff
@@ -523,6 +523,29 @@ impl Client {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns errors from the middleware, http backend, and network sockets.
+    ///
+    /// # Examples
+    /// ```no_run
+    /// # #[async_std::main]
+    /// # async fn main() -> surf::Result<()> {
+    /// let client = surf::client();
```
Fishrock123, line 540:
```suggestion
    
    /// use surf::Method;
    
```

Fishrock123 (changes requested):
Looks good but needs a small change to fix the doctest

src/client.rs
utsavm9: `cargo test` now passes locally after this change. I forgot to run it before making the PR, sorry about that!

Fishrock123 (approved):

Issue #277, opened 2021-01-14
TCP_NODELAY


robsmith11:
I don't see any way to set TCP_NODELAY (when using linux).  Is it not possible?

Fishrock123:
short answer: yes, at least for the curl/isahc client.

long answer: you're going to need to piece it together

------

By default Surf uses `IsahcClient` i.e. libcurl.

You'll need to create one from an `isahc::Client`: https://github.com/http-rs/http-client/blob/83b7fb0c1c55051362de22f389be49b65ca85754/src/isahc.rs#L24-L27

Which you'll need to build via `isahc::HttpClientBuilder` and use methods from the `isahc::config::Configurable` trait: https://docs.rs/isahc/0.9.10/isahc/struct.HttpClientBuilder.html#impl-Configurable

And the specific setting you'd be looking for is https://docs.rs/isahc/1.0.3/isahc/config/trait.Configurable.html#method.tcp_nodelay

robsmith11:
Thanks!  I got a bit lost with different layers of abstraction..

So it's not possible to call the config methods on the static default client?  Also, I saw that "h1" might become the default http client.  Is it possible to set TCP_NODELAY there too?

Fishrock123:
> So it's not possible to call the config methods on the static default client?

It is not possible to configure the default client at this time.

> Is it possible to set TCP_NODELAY there _(h1 client)_ too?

Not at present. [async-std does support nodelay](https://docs.rs/async-std/1.9.0/async_std/net/struct.TcpStream.html#method.set_nodelay), so it would have to be added to this part of http-client:
https://github.com/http-rs/http-client/blob/a9928a05732bdbcb33551de1ee2d37f60a43b44f/src/h1.rs#L59-L70

However, http-client currently does not have any way of specify options, see this issue for the beginning of that discussion: https://github.com/http-rs/http-client/issues/61

Fishrock123:
WIP draft PR up for the API in http-client: https://github.com/http-rs/http-client/pull/84

Fishrock123:
Available in https://github.com/http-rs/surf/releases/tag/v2.3.0

Issue #274, opened 2021-01-06
Please add support for setting request timeout


stevepryde:
I'm aware of the workaround in #267 but would really like a more ergonomic solution to set the request timeout in surf if possible. 

That workaround requires juggling both the `http-client` and `isahc` version to make sure the types line up between them and it feels really brittle. For example http-client v6.2 depends on isahc v0.9, whereas the latest version of isahc is currently 1.0. I wanted to use 1.0 but cannot because of the incompatibility.

Please add the ability to set the request timeout when creating a client, as well as the ability to set/override the request timeout per-request if desired. Thanks.

Fishrock123:
This needs work in `http-client`, which currently does not have any support for options. See https://github.com/http-rs/http-client/issues/61

Shnatsel:
I believe it's currently possible to set a timeout for the entire request via [`future::timeout`](https://docs.rs/async-std/1.9.0/async_std/future/fn.timeout.htm). That's great! However, the following timeouts are still missing:

1. Connect timeout (the server has to reply for the first time within a certain timeframe)
2. Read timeout (connection to be dropped after a certain period of inactivity)

The defaults for these values are rather high and it's often needed to override them. For example, the default read timeout is 5 minutes (actually enforced on the TCP level).

Fishrock123:
WIP draft PR up for the API in http-client: https://github.com/http-rs/http-client/pull/84

Fishrock123:
Also, please be aware this is possible externally with [`async_std::io::timeout()`](https://docs.rs/async-std/1.9.0/async_std/io/fn.timeout.html)!

fiag:
to customize isahc::HtttpClient which has a timeout option.
```
    let client = isahc::HttpClient::builder()
        .tcp_keepalive(Duration::from_secs_f32(3600.0))
        .tcp_nodelay()
        .timeout(Duration::from_secs_f32(10.0)
        .build()?;
    let client = http_client::isahc::IsahcClient::from_client(client);
    let client = surf::Client::with_http_client(client);
    client.get(....)
}

```

stevepryde:
> to customize isahc::HtttpClient which has a timeout option.
> 
> ```
>     let client = isahc::HttpClient::builder()
>         .tcp_keepalive(Duration::from_secs_f32(3600.0))
>         .tcp_nodelay()
>         .timeout(Duration::from_secs_f32(10.0)
>         .build()?;
[...]

Issue #255, opened 2020-10-22
base_url "/" handling


Fishrock123:
Taken from https://github.com/http-rs/surf/pull/252#issuecomment-707942927

`base_url` acts in unexpected ways when a trailing slash is missing. See that PR for more details.

ericsampson:
there's a spec for this, [RFC](https://datatracker.ietf.org/doc/html/rfc3986)

So AFAIC what's described in #252 as a "bug" is actually correct behavior by the underlying library.

For example, the .NET HttpClient documentation has this note:
> Note that all characters after the right-most "/" in the base URI are excluded when combined with the message URI. See RFC 3986 Uniform Resource Identifier (URI) Generic Syntax specification.

PR #310, opened 2021-07-05
feat: add Config for surf

Fishrock123:
Finally adds top-level configuration to surf, built on top of http-client's still as of yet unstable config.

Related to https://github.com/http-rs/http-client/pull/86

Closes https://github.com/http-rs/surf/issues/274
Closes https://github.com/http-rs/surf/issues/277

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,0 +1,185 @@
+//! Configuration for `HttpClient`s.
+
+use std::sync::Arc;
+use std::{collections::HashMap, fmt::Debug, time::Duration};
+
+use http_client::{Config as HttpConfig, HttpClient};
+use http_types::headers::{HeaderName, HeaderValues, ToHeaderValues};
+
+use crate::http::Url;
+use crate::Result;
+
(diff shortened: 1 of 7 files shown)


src/config.rs
```diff
@@ -0,0 +1,100 @@
...
+
+/// Configuration for `surf::Client`s and their underlying HTTP clients.
+#[non_exhaustive]
+#[derive(Clone, Debug)]
+pub struct Config {
+    /// The base URL for a client. All request URLs will be relative to this URL.
+    ///
+    /// Note: a trailing slash is significant.
+    /// Without it, the last path component is considered to be a “file” name
+    /// to be removed to get at the “directory” that is used as the base.
```
jbr, line 20: Could we just add a slash at the end for them and remove this footgun? Is there ever a case where we wouldn't want that?

Fishrock123: Probably? This issue is tracked at https://github.com/http-rs/surf/issues/255

Issue #315, opened 2021-07-16
wasm feature flags & getrandom dependency


ufoscout:
I created a new empty cargo project with this `Cargo.toml` file:
```toml
[package]
name = "test_surf_wasm"
version = "0.1.0"
edition = "2018"

[dependencies]
surf = { version = "2.2", default-features = false, features = ["wasm-client"] }
```

Trying to compile it with target `wasm32-unknown-unknown` fails with:
```bash
$ cargo build --target wasm32-unknown-unknown
    Updating crates.io index
   Compiling version_check v0.9.3
   Compiling proc-macro2 v1.0.27
   Compiling unicode-xid v0.2.2
   Compiling syn v1.0.73
   Compiling cfg-if v1.0.0
   Compiling typenum v1.13.0
   Compiling log v0.4.14
   Compiling serde v1.0.126
   Compiling wasm-bindgen-shared v0.2.74
   Compiling serde_derive v1.0.126
   Compiling lazy_static v1.4.0
   Compiling bumpalo v3.7.0
   Compiling ryu v1.0.5
   Compiling wasm-bindgen v0.2.74
   Compiling proc-macro-hack v0.5.19
   Compiling futures-core v0.3.15
   Compiling subtle v2.4.1
   Compiling serde_json v1.0.64
   Compiling opaque-debug v0.3.0
   Compiling futures-channel v0.3.15
   Compiling autocfg v1.0.1
   Compiling memchr v2.4.0
   Compiling semver-parser v0.7.0
   Compiling futures-sink v0.3.15
   Compiling matches v0.1.8
   Compiling libc v0.2.98
   Compiling futures-io v0.3.15
   Compiling stdweb-internal-runtime v0.1.5
   Compiling itoa v0.4.7
   Compiling percent-encoding v2.1.0
   Compiling ppv-lite86 v0.2.10
   Compiling getrandom v0.1.16
   Compiling pin-project-lite v0.2.7
   Compiling sha1 v0.6.0
   Compiling tinyvec_macros v0.1.0
   Compiling proc-macro-nested v0.1.7
   Compiling slab v0.4.3
   Compiling futures-task v0.3.15
   Compiling pin-utils v0.1.0
   Compiling base-x v0.2.8
   Compiling const_fn v0.4.8
   Compiling cache-padded v1.1.1
   Compiling discard v1.0.4
   Compiling crossbeam-utils v0.8.5
   Compiling event-listener v2.5.1
   Compiling anyhow v1.0.42
[...]

PR #316, opened 2021-07-27
fix: pin getrandom to work on wasm

jkelleyrtp:
This PR pins getrandom to the "js" feature so that surf compiles on wasm32 again.

Solves #315 

Before this change, when I tried to build the repo for wasm:
```
cargo build --target wasm32-unknown-unknown --features wasm-client --no-default-features

   Compiling getrandom v0.2.3
error: the wasm32-unknown-unknown target is not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support
   --> /Users/jonkelley/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.3/src/lib.rs:219:9
    |
219 | /         compile_error!("the wasm32-unknown-unknown target is not supported by \
220 | |                         default, you may need to enable the \"js\" feature. \
221 | |                         For more information see: \
222 | |                         https://docs.rs/getrandom/#webassembly-support");
    | |_________________________________________________________________________^

error[E0433]: failed to resolve: use of undeclared crate or module `imp`
   --> /Users/jonkelley/.cargo/registry/src/github.com-1ecc6299db9ec823/getrandom-0.2.3/src/lib.rs:246:5
    |
246 |     imp::getrandom_inner(dest)
    |     ^^^ use of undeclared crate or module `imp`

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0433`.
error: could not compile `getrandom`

To learn more, run the command again with --verbose.

```

With these feature flags, I can build properly:

```
cargo build --target wasm32-unknown-unknown --features wasm-client --no-default-features

    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,10 @@ cfg-if = "1.0.0"
 # encoding
 encoding_rs = { version = "0.8.20", optional = true }
 
-# wasm-client
+# wasm-client config
+[target.'cfg(target_arch = "wasm32")'.dependencies.getrandom]
+version = "0.2"
+features = ["js"]
 [target.'cfg(target_arch = "wasm32")'.dependencies.web-sys]
 version = "0.3.25"
 optional = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,12 @@ hyper-client = [
     "default-client",
     "async-std/tokio02",
 ]
-wasm-client = ["http-client/wasm_client", "default-client"]
+wasm-client = [
+    "http-client/wasm_client",
+    "default-client",
+    "getrandom/js",
+    "web-sys",
+]
 default-client = []
 middleware-logger = []
 # requires web-sys for TextDecoder on wasm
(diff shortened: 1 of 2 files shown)


CryZe:
Keep in mind that the approach in this PR requires "target specific features", which means that this change is somewhat incompatible with `resolver = 1`. This may be fine, but it's something worth considering.

jkelleyrtp:
> Keep in mind that the approach in this PR requires "target specific features", which means that this change is somewhat incompatible with `resolver = 1`. This may be fine, but it's something worth considering. It probably makes more sense to tie it to the `wasm-client` feature.

Thanks for the pointers. I switched the flag selection to use normal feature flags instead of the target-dependent flags.

Fishrock123 (approved):

Fishrock123:
I'll try to get 2.3.0 out "soon".

CryZe:
web-sys, getrandom and encoding now always get pulled in as dependencies. Target specific crates are not the problem. This part should probably be reverted.

Fishrock123:
Those should only be pulled in when `wasm-client` is selected though?

CryZe:
Yes, but due to resolver = 1, there are technically situations where that feature is active, but you aren't even compiling for WASM. (Resolver 1 is truly a mess)

Fishrock123:
Witch situations would those be? You have to manually activate the feature or have something else consciously activate it.

CryZe:
I believe it can be if you have multiple final binaries in the same workspace, where one ends up activating the feature for WASM and one compiles for x86. Another alternative is that there might be a crate that depends on surf, but always expects to be used only on WASM, so that crate might end up in your dependency graph (since all targets are always in the dependency graph, you can often find fuchsia and redox specific crates in Cargo.lock) and then even if you never intend to compile for WASM, this feature will be active. That's the issue with resolver 1. Technically you can ignore resolver 1 and say that surf is only compatible with resolver 2 (which basically has separate dependency graphs when it comes to the features for each target), but this is trivially fixable by undoing a few lines of this PR.

PR #333, opened 2021-12-05
Depend on http-types with only the "fs" feature, no other default features

joshtriplett:
This allows users of surf to avoid the substantial dependency tree for
the "cookie" feature if they don't need it.


Fishrock123:
is this semver-major?

joshtriplett:
@Fishrock123 I don't think so, no. Dropping "fs" would be semver-major, but I think as long as we keep that, the rest would be fine.

PR #337, opened 2022-01-13
Allow building with `--all-features`

misuzu:
`h1-client-rustls` is preferred when `h1-client` is also enabled.

I just want to use `h1-client-rustls` with sentry: https://github.com/getsentry/sentry-rust/pull/404
See https://github.com/rust-lang/cargo/issues/4328 and https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/features.md#feature-unification


Fishrock123:
`--all-features` does not work with Surf because it pulls in multiple conflicting defaults.

misuzu:
> `--all-features` does not work with Surf because it pulls in multiple conflicting defaults.

With this change, they aren't.

PR #340, opened 2022-02-19
Upgrade `rustls` to 0.19

jakubadamw:
The 0.20 upgrade of would have been much more substantial with quite a few breaking changes, so this is not included.


mfelsche:
Turns out we depend on this change with our usage of `surf` and `http-client` and would love to get this out with a new release. :heart:

jakubadamw:
@Fishrock123 any chance we could get this in?

halvors:
@Fishrock123 Bump?

Fishrock123:
since this crate exposes rustls api this is a breaking change, requiring a major, and that makes it require effort and I am short on time these days

PR #343, opened 2022-05-09
Support requiring requests to go to the configured base URL's origin

joshtriplett:
Add `Config::set_mandatory_base_origin()` to enable this behavior.
Useful if the `Config` includes authentication headers or similar.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://book.async.rs/overview
 
 ## [Unreleased]
 
+### Additions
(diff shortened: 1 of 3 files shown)


Fishrock123:
Should this be a cargo option? Or do you think it's better on `Config` like you have in this PR?

joshtriplett:
@Fishrock123 I definitely don't think it should be a cargo feature; that'd be backwards-incompatible for anyone not expecting that behavior, including other users elsewhere in the dependency tree that didn't set that feature.

I think it makes sense to set this on a config-by-config basis.

PR #349, opened 2022-08-08
Update README example 

shubhamkumar13:
So this might be a very small change but I was confused how to use the previous version of this example.

This change makes the example plug and play and easier for beginners like me to use it.

Would love to have an opinion on this :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,10 +61,8 @@ Whether it's a quick script, or a cross-platform SDK, Surf will make it work.
 
 ## Examples
 ```rust
-use http_types;
-
 #[async_std::main]
-async fn main() -> Result<(), http_types::Error> {
+async fn main() -> Result<(), surf::Error> {
     let mut res = surf::get("https://httpbin.org/get").await?;
     dbg!(res.body_string().await?);
     Ok(())
@@ -73,10 +71,8 @@ async fn main() -> Result<(), http_types::Error> {
 
 It's also possible to skip the intermediate `Response`, and access the response type directly.
 ```rust
-use http_types;
-
 #[async_std::main]
-async fn main() -> Result<(), http_types::Error> {
+async fn main() -> Result<(), surf::Error> {
     dbg!(surf::get("https://httpbin.org/get").recv_string().await?);
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


Fishrock123 (approved):

jbr:
Is there any reason this requires a dependency on http_types instead of using surf::Error?

shubhamkumar13:
I think @jbr you are right there is no specific reason for http_types.
I didn't realise there is a `surf::Error`. 

So initially I was using the [hello_world.rs](https://github.com/http-rs/surf/blob/main/examples/hello_world.rs) example and I landed on using the exterior part of the main function :
```rs
#[async_std::main]
async fn main() -> Result<(), http_types::Error> {
    // fill the readme stuff here
}
```

And fill the above function with the code mentioned in the [README.md](https://github.com/http-rs/surf#examples)
which ended up looking like this :
```rs
#[async_std::main]
async fn main() -> Result<(), http_types::Error> {
    let mut res = surf::get("https://httpbin.org/get").await?;
    dbg!(res.body_string().await?);
    Ok(())
}
```
which resulted in the following error
```bash
❯ cargo run
   Compiling trying_surf v0.1.0 (/home/sk/trying_surf)
error[E0433]: failed to resolve: use of undeclared crate or module `http_types`
 --> src/main.rs:9:31
  |
9 | async fn main() -> Result<(), http_types::Error> {
  |                               ^^^^^^^^^^ use of undeclared crate or module `http_types`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `trying_surf` due to previous error
```
so adding a `use http_types::Error;` fixed the above statement and I went with it.

But now that you mentioned it when I change the code to :
```rs
#[async_std::main]
async fn main() -> Result<(), surf::Error> {
    let mut res = surf::get("https://httpbin.org/get").await?;
    dbg!(res.body_string().await?);
    Ok(())
}
```
it looks good! and runs without any errors!!

So I'll update the PR and if it's ok can you approve it? Thanks for the pointer :smiley:

shubhamkumar13:
is it ok if I update the examples directory it can be a bit confusing for beginners?

PR #365, opened 2025-08-13
fix `cargo test` in current version of rustc

alycda:
error: unexpected `cfg` condition value: `docs`

note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration


alycda:
cc @yoshuawuyts @Fishrock123 can I get some eyes on this please? Thanks in advance

